PR #2, opened 2024-07-02
Maya: USD workflow with maya usd

BigRoy:
> [!NOTE] 
> PR moved from https://github.com/ynput/ayon-core/pull/334

## Changelog Description

Implement more of Maya USD workflows:

- Add more options to the Maya USD Extractor
- Export a layer of a Maya USD proxy, e.g. lookdev done with Maya's lookdevx.
- Prototype loading of USD references into a Maya USD proxy while keeping it managed by the pipeline
- Prototype loading of Maya references into a Maya USD proxy while keeping it managed by the pipeline

## Additional info

Some of the logic here has been used in our production - but there's also quite a bit that we didn't use at all in the actual production. The logic here can thus be considered mostly a prototype and up for discussion.

We mostly used the loading of USD files to a proxy - for previewing purposes. And the Maya USD export from Maya.

I also had [this prototype](https://gist.github.com/BigRoy/7d73ec1bfcc5944a6aa887ee31b1bf44) to export native maya model + lookdev done with arnold to just a "look" layer override for a USD stage. (So basically a look export without the geometry present so that you could overlay it over the asset; however we did not use that.)

Anyway, input is more than welcome!

This is basically the Maya part of this equivalent OpenPype PR https://github.com/ynput/OpenPype/pull/5925 but with some more improvements we did while using some parts in production.

## Testing notes:

1. For now, play around with the features! (Until PR is ready and out of draft!)


BigRoy:
Closing this in favor of:
- #59
- #60 
- #61

---

> I also had [this prototype](https://gist.github.com/BigRoy/7d73ec1bfcc5944a6aa887ee31b1bf44) to export native maya model + lookdev done with arnold to just a "look" layer override for a USD stage. (So basically a look export without the geometry present so that you could overlay it over the asset; however we did not use that.)

This, mentioned in the PR description isn't in this current PR but also isn't separated or added into one of the other PRs.

PR #5, opened 2024-07-03
Maya: Workfile testing

moonyuet:
### Changelog Description
Ported from https://github.com/ynput/ayon-core/pull/254


moonyuet:
@BigRoy @antirotor Do we still need to continue with this PR or we have any other solution for that?

antirotor:
I'd say it is sill worth the effort but we need to rethink how this is executed and where. This should be part of the bigger testing suite to be executed on site to test the whole pipeline. There is a value in ability to run it on demand just in Maya but also to run it as a part of bigger automated testing run. Before additional work here, we should have a little brainstorm how to align it.

PR #12, opened 2024-07-08
Maya: JSON layout for Unreal using local transform matrix instead of world matrix

moonyuet:
### Changelog Description
When extracting layout, the json data should align with what Unreal used to load for the maya presets.
You can find the screengrab of the loaded preset for abc here:
https://github.com/ynput/ayon-unreal/issues/15#issuecomment-2213039569

### Additional Info
Please test along with this PR https://github.com/ynput/ayon-unreal/pull/35

### Testing Notes
1. Launch Maya
2. Create Layout with loaded assets
3. Publish
4. Load the layout back to the related loaded assets.


BigRoy:
Don't have access to unreal setup on my end and it's hard to tell whether this is the correct fix. It is somewhat odd to me that just above we are storing translate, rotate, scale into the JSON file but store the matrix in an altered way.

Anyway, the code change itself seems fine. Whether it resolves an issue, I'm not sure.

moonyuet:
> Don't have access to unreal setup on my end and it's hard to tell whether this is the correct fix. It is somewhat odd to me that just above we are storing translate, rotate, scale into the JSON file but store the matrix in an altered way.
> 
> Anyway, the code change itself seems fine. Whether it resolves an issue, I'm not sure.

it also looks odd to me too ngl. But what I have been seen in the unreal settings seems to be that the layout setting doesn't align with the presets set in the unreal maya preset, so that's why some transformations mess up. It would be great to know whether this extract layout has been used anywhere other than Unreal or it has been used in other extensions as I do change something to be suitable for just loading abc in unreal

LiborBatek (commented):
Once I have produced ```layout``` product in ```maya``` and loaded into UE as ```load JSON layout``` I got my layout with applied ```transforms``` as when set for ```maya Alembic to UE``` but seems wrong oriented anyway... as seen on first img (my room being rotated to side instead of floor being the ground level) - note those ```transforms``` and its values...

Screenshot 2024-07-08 141743

once reset it to ```zero transform``` it looks much sensible aka correctly oriented in space as floor being ground level (wether its good or bad I cant tell but obviously closer to normal behaviour for sure)


Screenshot 2024-07-08 141844

and as seen the interior  from FPView:


Screenshot 2024-07-08 142029

LiborBatek (commented):
Now the ```RotX``` being set to ```90``` and still not correct?
Screenshot 2024-07-08 173646

moonyuet:
> Now the `RotX` being set to `90` and still not correct? Screenshot 2024-07-08 173646

The updated commit should import layout as if this one.
image

LiborBatek:
ok I will do another test round and remove all assets (as some can be already existing with wrong transforms??) outside of the layout which re-uses such asset.

will update here asap, thx!

LiborBatek:
@moonyuet it seems working when ```FBX``` assets been loaded into ```layout``` ...not sure if the integration first checks if ```FBX``` repre exists but it did for model of the environment (as FBX exists for that one), but the rest of the assets in ```alembic``` repre seems not correct - I guess due to the fact we import those with wrong transforms?? not necessirely when ```layout``` product been loaded but maybe as they have been imported aka loaded in advance using AYON Loader (wrong transforms by AYON ??)

Another issue being that if any assets do not exists already within UE project (aka been loaded in advance) the ```Load Layout``` will fail resulting into empty UE map. Again I dont know if any particular UE addon branch needed in companion to this PR?? to make that working...

moonyuet:
> @moonyuet it seems working when `FBX` assets been loaded into `layout` ...not sure if the integration first checks if `FBX` repre exists but it did for model of the environment (as FBX exists for that one), but the rest of the assets in `alembic` repre seems not correct - I guess due to the fact we import those with wrong transforms?? not necessirely when `layout` product been loaded but maybe as they have been imported aka loaded in advance using AYON Loader (wrong transforms by AYON ??)
> 
We need to have the option which allows users to export layout for alembic and fbx separately. 

> Another issue being that if any assets do not exists already within UE project (aka been loaded in advance) the `Load Layout` will fail resulting into empty UE map. Again I dont know if any particular UE addon branch needed in companion to this PR?? to make that working...

This should be issue for the layout loader in unreal. I will make an issue for that.

moonyuet:
@LiborBatek The updated PR can successfully flip y axis to z axis. The only circumstance is not working when the y rotation and z rotation both filled out with different values
Should we make a validator to make sure these y and z rotations are not filled at the same time to ensure it works correctly in unreal?

LiborBatek:
> @LiborBatek The updated PR can successfully flip y axis to z axis. The only circumstance is not working when the y rotation and z rotation both filled out with different values Should we make a validator to make sure these y and z rotations are not filled at the same time to ensure it works correctly in unreal?

thats sounds really restrictive...as things living in 3 dimensional space, no?? I guess we have to solve it somehow...maybe someone can step in and help you out?

moonyuet:
> > @LiborBatek The updated PR can successfully flip y axis to z axis. The only circumstance is not working when the y rotation and z rotation both filled out with different values Should we make a validator to make sure these y and z rotations are not filled at the same time to ensure it works correctly in unreal?
> 
> thats sounds really restrictive...as things living in 3 dimensional space, no?? I guess we have to solve it somehow...maybe someone can step in and help you out?

Yes It would be great if @antirotor or @BigRoy or even some potential community members can join to discuss on that and needs more research or discussion on figuring out as when y axis moves and it would move the z axis and vice versa. The numbers are varied under the condition of z-axis being over 90, and y axis being less than 90. Anyways, it would be great to have some discussion on how to fix the rotation when it has Y axis and Z axis.
We can test on this one or create another issue to do the decent fix.
image

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -112,10 +111,12 @@ def process(self, instance):
...
             t = om.MVector(t.x, t.z, -t.y)
+            s = transform.scale(om.MSpace.kObject)
+            s = [s[0], s[2], s[1]]
+            r = transform.rotation()
             transform.setTranslation(t, om.MSpace.kWorld)
-            transform.rotateBy(
-                om.MEulerRotation(math.radians(-90), 0, 0), om.MSpace.kWorld)
-            transform.scaleBy([1.0, 1.0, -1.0], om.MSpace.kObject)
+            transform.setRotation(om.MEulerRotation(r.x, r.z, r.y))
+            transform.setScale(s, om.MSpace.kObject)
```
BigRoy, line 119: What is it that we're trying to do - are we trying to rotate an object in world-space around the pivot point?
Isn't that just a matrix multiplication?

moonyuet: As maya by default is Y-axis up and unreal is Z-axis up, what we are trying to do is make sure transformation data which are Y-axis up converting to the data which are Z-axis up. Would be helpful if you can come up with some hints(or codes) for the matrix multiplication. We can edit the identity matrix from the basis, of which we dont need to setScale and setTranslation like this.  But for the rotation is more complicated issue, it really needs some maths to be done.

moonyuet:
@LiborBatek 
the current rotation should be correct for anything less than 90 degrees
it should be "correct" too for greater than 90 degrees, but you need to double check on the placement of the object as it looks a bit different.

antirotor:
I think the solution is hidden here:

https://github.com/Autodesk/LiveLink/blob/98f230e7333aae5a70c281ddbfe13ac090692f86/Source/Programs/MayaUnrealLiveLinkPlugin/MayaUnrealLiveLinkUtils.cpp#L117-L159

That function is converting maya transform to unreal transform. First time I am glad OpenMaya API is more C++ then Python, because rewritting this to Python should be easy. I'll test it tomorrow.

moonyuet:
> I think the solution is hidden here:
> 
> https://github.com/Autodesk/LiveLink/blob/98f230e7333aae5a70c281ddbfe13ac090692f86/Source/Programs/MayaUnrealLiveLinkPlugin/MayaUnrealLiveLinkUtils.cpp#L117-L159
> 
> That function is converting maya transform to unreal transform. First time I am glad OpenMaya API is more C++ then Python, because rewritting this to Python should be easy. I'll test it tomorrow.

I've converted the script and will test to see if it matches the axis in unreal
You can find the example snippet below:
```python
from maya.api import OpenMaya as om
import math
row_length = 4
t_matrix_list = cmds.xform("pCube1", query=True, matrix=True)

maya_transform_mm = om.MMatrix(t_matrix_list)
ue_transform_mm = om.MMatrix()
for i in range(0, row_length):
	first_row = maya_transform_mm.getElement(i, 0)
	second_row = maya_transform_mm.getElement(i, 1)
	third_row = maya_transform_mm.getElement(i, 2)
	fourth_row = maya_transform_mm.getElement(i, 3)
	if i == 1:
		ue_transform_mm.setElement(i, 0, -first_row)
		ue_transform_mm.setElement(i, 1, second_row)
		ue_transform_mm.setElement(i, 2, -third_row)
		ue_transform_mm.setElement(i, 3, -fourth_row)
	else:
		ue_transform_mm.setElement(i, 0, first_row)
		ue_transform_mm.setElement(i, 1, -second_row)
		ue_transform_mm.setElement(i, 2, third_row)
		ue_transform_mm.setElement(i, 3, fourth_row)

transform = om.MTransformationMatrix(ue_transform_mm)
rotation = transform.rotation().asQuaternion()
transform.setRotation(om.MQuaternion(rotation.x, rotation.y, rotation.z, rotation.w))

translate = transform.translation(om.MSpace.kWorld)
scale = transform.scale(om.MSpace.kWorld)
transform.setTranslation(om.MVector(translate.x, translate.y, translate.z), om.MSpace.kWorld)
transform.setScale([scale[0], scale[1], scale[2]], om.MSpace.kWorld)

t_matrix_list = list(transform.asMatrix())
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -106,7 +106,7 @@ def process(self, instance):
             json_path = os.path.join(stagingdir, json_filename)
             t_matrix = []
             with lib.maintained_selection():
-                cmds.ls(asset, dagObjects=True)
+                cmds.select(asset, noExpand=True)
                 sel = om.MGlobal.getActiveSelectionList()
                 dagpath = sel.getDependNode(0)
```
BigRoy, line 111: This may be even better as:
```suggestion
                sel = om.MSelectionList()
                sel.add(asset)
                dagpath = sel.getDependNode(0)
```

That way you don't even need to make any selection in the scenes and you can also remove the maintained selection.

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -82,56 +81,53 @@ def process(self, instance):
...
+                dagpath = sel.getDependNode(0)
+                ue_transform = om.MFnTransform(dagpath)
+                with lib.maintained_transformation(ue_transform):
+                    # make sure the data doesn't change during context
+                    convert_transform = om.MTransformationMatrix(convert_transform_mm)
+                    final_ue_transform = ue_transform.setTransformation(convert_transform)
+                    t_matrix_list = list(final_ue_transform.transformation().asMatrix())
+
+                    for i in range(0, len(t_matrix_list), row_length):
+                        t_matrix.append(t_matrix_list[i:i + row_length])
```
BigRoy, line 120: Avoiding the selection may be better?
```suggestion
            sel = om.MSelectionList()
            sel.add(asset)
            dagpath = sel.getDagPath(0)
            ue_transform = om.MFnTransform(dagpath)
            with lib.maintained_transformation(ue_transform):
                # make sure the data doesn't change during context
                convert_transform = om.MTransformationMatrix(convert_transform_mm)
                final_ue_transform = ue_transform.setTransformation(convert_transform)
                t_matrix_list = list(final_ue_transform.transformation().asMatrix())
                for i in range(0, len(t_matrix_list), row_length):
                    t_matrix.append(t_matrix_list[i:i + row_length])
```

I'm not sure what that matrix logic is trying to do... we're using `setTransformation` to set a matrix to then after get that transformation matrix.

BigRoy: As far as I can see - this should produce the exact same result:

```suggestion
            convert_transform_mm = list(convert_transform_mm)
            t_matrix_list = []
            for i in range(0, len(t_matrix_list), row_length):
                t_matrix.append(t_matrix_list[i:i + row_length])
```

antirotor:
we are getting there - orientations are now good, but it is shifting translation in wrong axes

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -1,13 +1,85 @@
...
+def get_converted_matrix(matrix: Union[om.MMatrix, list]) -> om.MMatrix:
+    """Get Maya transform converted to Unreal Engine transform.
+    Perform all conversion steps similar to:
+        https://github.com/Autodesk/LiveLink/blob/main/Source/Programs/MayaUnrealLiveLinkPlugin/Subjects/MLiveLinkPropSubject.cpp#L122-L129  # noqa
+    """
+    matrix = om.MMatrix(matrix)
+    matrix = rotate_coordinates_system_for_unreal(matrix)
+    matrix = build_ue_transform_from_maya_transform(matrix)
+    if om.MGlobal.isYAxisUp():
+        rot = om.EulerRotation(0, 0, math.radians(90)).asMatrix()
```
moonyuet, line 78:
```suggestion
        rot = om.MEulerRotation(0, 0, math.radians(90)).asMatrix()
```

BigRoy:
So with recent commits the transformation should be similar to what the Unreal Live Link plug-in does, e.g [here](https://github.com/Autodesk/LiveLink/blob/main/Source/Programs/MayaUnrealLiveLinkPlugin/Subjects/MLiveLinkPropSubject.cpp#L122-L129).

Unfortunately I don't have a running Unreal instance to really investigate the behavior of the rotation differences and get into the nitty gritty - I've just refactored the code to restructure what it's doing and make it clear why it's doing something (also simplifying a lot of logic) whilst documenting how we're matching the Unreal Live Link conversion logic.

However, what I understood from @moonyuet this still didn't provide us the right results. I understood there's also "some" additiona logic in the Load Layout loader in Unreal that may also be doing something special with the json data it's receiving from here - which would be [around this code here](https://github.com/ynput/ayon-unreal/blob/0ece578f9db9b51d16f0328786985fd4c79e5283/client/ayon_unreal/plugins/load/load_layout.py#L103-L136).

However, now with the basis set to identity matrix that should adhere closely to the matrix in the data since it'd just make it the transformation matrix being passed in.

moonyuet:
> So with recent commits the transformation should be similar to what the Unreal Live Link plug-in does, e.g [here](https://github.com/Autodesk/LiveLink/blob/main/Source/Programs/MayaUnrealLiveLinkPlugin/Subjects/MLiveLinkPropSubject.cpp#L122-L129).
> 
> Unfortunately I don't have a running Unreal instance to really investigate the behavior of the rotation differences and get into the nitty gritty - I've just refactored the code to restructure what it's doing and make it clear why it's doing something (also simplifying a lot of logic) whilst documenting how we're matching the Unreal Live Link conversion logic.
> 
> However, what I understood from @moonyuet this still didn't provide us the right results. I understood there's also "some" additiona logic in the Load Layout loader in Unreal that may also be doing something special with the json data it's receiving from here - which would be [around this code here](https://github.com/ynput/ayon-unreal/blob/0ece578f9db9b51d16f0328786985fd4c79e5283/client/ayon_unreal/plugins/load/load_layout.py#L103-L136).
> 
> However, now with the basis set to identity matrix that should adhere closely to the matrix in the data since it'd just make it the transformation matrix being passed in.
> 
> @antirotor Any chance you're able to give the current state a test run - and from your POV can best elaborate on how you feel the rotation is misbehaving (and/or anything else is misbehaving.)

Good to note that the translate and scale are good with basis matrix we have discussed
The best result for rotations is nothing changed.
```
            json_element["basis"] = [
                [1, 0, 0, 0],
                [0, 0, 1, 0],
                [0, 1, 0, 0],
                [0, 0, 0, 1]
            ]

```
But I think it is probably because xyz rotation has huge differences value than xzy rotation in Quadernotions
rot = cmds.xform("asset_modelMain_01_:_GRP", query=True, rotation=True)
```
from maya.api import OpenMaya
print(OpenMaya.MEulerRotation(rot[0],rot[1], rot[2]).asQuaternion())
# (0.719786, -0.128318, 0.648645, 0.211428)
print(OpenMaya.MEulerRotation(rot[0],rot[2], rot[1]).asQuaternion())
#(0.719786, -0.405357, -0.52239, 0.211428)

```
https://github.com/ynput/ayon-maya/blob/8ecbc75a68d46973cfff02b0f10636180710eef1/client/ayon_maya/plugins/publish/extract_layout.py#L69-L80

BigRoy:
Yup - I'll leave it as is - feel free to revert commits or change things around again. Hope the restructure of the code at least organizes it better for readability. ❤️ 

Sorry for not being able to define a decisive answer.

moonyuet:
Just want to update this: I have tested with unreal where I put the geometry more than 90 degree in rotZ in Maya(RotationXYZ: 30, 90,120) , and the geometry gets into unreal, it went correct when i do comparison with the geometry in the same rotation(Rotation XZY: 30, 120, 90)
Similar rules applied to unity, other game engine. 
I find this post helpful for this case: https://forum.unity.com/threads/rotation-woes-going-past-90deg-on-any-axis-forces-others-to-180-or-180-why.69224/?fbclid=IwZXh0bgNhZW0CMTAAAR0-QXsvL25Zte5bIeZ8izedcOjap2tkR7sl-ThFAhXQSat6L6NTg2OtnT0_aem_qeWCUBChuZCLymTsjJtBlg
there are multiple correct ways of representing the quaternion as euler angles. (0, 0, -90) is exactly the same thing as (180, 180, 90)

moonyuet:
The attached video to explain how the rotation matches even with different numbers, as there are multiple representations of euler rotation in reality.
https://github.com/user-attachments/assets/a2ac7282-5c06-4c00-83cd-b40c586d2d49

BigRoy:
> The attached video to explain how the rotation matches even with different numbers, as there are multiple representations of euler rotation in reality. https://github.com/user-attachments/assets/a2ac7282-5c06-4c00-83cd-b40c586d2d49

Admittedly, without explanation over the video it's hard to really understand what difference is being highlighted in the video. Not super problematic, but it does somewhat make the video hard to reference.

You're manually setting values - but those appear the exact same as the loader does or the other one is already? I'm not seeing the "differing" values?

moonyuet:
> > The attached video to explain how the rotation matches even with different numbers, as there are multiple representations of euler rotation in reality. https://github.com/user-attachments/assets/a2ac7282-5c06-4c00-83cd-b40c586d2d49
> 
> Admittedly, without explanation over the video it's hard to really understand what difference is being highlighted in the video. Not super problematic, but it does somewhat make the video hard to reference.
> 
> You're manually setting values - but those appear the exact same as the loader does or the other one is already? I'm not seeing the "differing" values?

I set some random values into the transformations and export these data as layout. And in the unreal engine scene, there are two models as you see in the video(one is with the data which containers the transformation data from load layout, the other is original object with zero translation and zero transform).
When I insert the values accordingly to the transformation from maya to the original object, and its transformation matches with the object which is with the data from the layout. So that proves the rotation are all same even if they are at different values.

BigRoy:
> So that proves the rotation are all same even if they are at different values.

Maybe I'm just blind and missed where the values were actually different between the two loaded entries in Unreal - but if that's clearly visible, then perfect 👍 ⭐

antirotor (changes requested):
image

the left most and the right most are wrong. Left one in Unreal is `X:0` `Y:0` `Z:-90` -> `Z` should +90 to look correct. Also the right most in Unreal is `X:0`  `Y:0` `Z:90` but the `Z` should be -90.

So I guess there is still issue of signs.

---
Note: in Maya, the left-most is 0, -90, 0 and the right one is 0, +90, 0

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -1,13 +1,65 @@
...
+    Returns:
+        List[om.MMatrix]: List of transformation matrix of the asset
+    """
+    convert_transform = om.MTransformationMatrix(transform_mm)
+
+    convert_tanslation = convert_transform.translation(om.MSpace.kWorld)
+    convert_tanslation = om.MVector(convert_tanslation.x, convert_tanslation.z, -convert_tanslation.y)
+    convert_scale = convert_transform.scale(om.MSpace.kObject)
+    convert_transform.setTranslation(convert_tanslation, om.MSpace.kWorld)
+    if math.radians(rotation[0]) <= 0 or math.radians(rotation[0]) == 0 and math.radians(rotation[1]) == 0:
```
BigRoy, line 48: The radians conversion here is redundant. We're comparing against 0 and the conversion of radians/degrees would make no difference to that. 

Also, the second check "looks redundant" because it appears included in the first check BUT instead due to `or`  versus `and` ordering I have a feeling that last `and` is merged with the second `or`. However, I'd still add parentheses then for clarity.
```suggestion
    if rotation[0] <= 0 or (rotation[0] == 0 and rotation[1] == 0):
```

Also, HOW did you come up with this solution. This feels overly "magical" and by pure chance just applying certain adjusted values. It smells like something that magically works for one case but will bite us for other cases. Would be great if a comment could be added explaining WHY this works so future us understand WHY it's done this way.

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -1,13 +1,65 @@
...
+    for i in range(0, len(value), 4):
+        result.append(list(value[i:i + 4]))
+    return result
+
+
+def convert_transformation_matrix(transform_mm: om.MMatrix, rotation: list) -> om.MMatrix:
+    """Convert matrix to list of transformation matrix for Unreal Engine import.
+
+    Args:
+        transform_mm (om.MMatrix): World Matrix for the asset
```
BigRoy, line 36: We are not actually passing it the world matrix but a local matrix?

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -1,13 +1,65 @@
...
+        rotation (list): Rotations of the asset
```
BigRoy, line 37: Why do we need to specify the rotation separately? The rotation should be in the `MMatrix` as well? Can't we just extract it from it? Why does it need to be a separate input argument?

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -1,13 +1,65 @@
...
+
+    Returns:
+        List[om.MMatrix]: List of transformation matrix of the asset
+    """
+    convert_transform = om.MTransformationMatrix(transform_mm)
+
+    convert_tanslation = convert_transform.translation(om.MSpace.kWorld)
+    convert_tanslation = om.MVector(convert_tanslation.x, convert_tanslation.z, -convert_tanslation.y)
```
BigRoy, line 45: Typo?
```suggestion
    convert_translation = convert_transform.translation(om.MSpace.kWorld)
    convert_translation = om.MVector(convert_translation.x, convert_translation.z, -convert_translation.y)
```

client/ayon_maya/plugins/publish/extract_layout.py
moonyuet: we need to readjust the rotation input from the cmds.xform(rotation=True), instead of using the one from the matrix, we are using the one from rotation. It's metathetically easier, if you have any other solution which is helpful for this case, feel free to suggest.

client/ayon_maya/plugins/publish/extract_layout.py
moonyuet: > Also, HOW did you come up with this solution. This feels overly "magical" and by pure chance just applying certain adjusted values. It smells like something that magically works for one case but will bite us for other cases. Would be great if a comment could be added explaining WHY this works so future us understand WHY it's done this way.

I hope I can explain why in certain way, but for certain case, it would be even less persuasive with the explanation. But I will try to.

client/ayon_maya/plugins/publish/extract_layout.py
BigRoy: Ah I see. You're using the raw rotation values from the node. I have a very odd feeling that this will break in many cases because rotation values of e.g. 360+ like 720 or 1080 which in essence are then also 0 rotation as a result may break in the same way in Unreal yet we're now not catching those?

moonyuet: > Ah I see. You're using the raw rotation values from the node. I have a very odd feeling that this will break in many cases because rotation values of e.g. 360+ like 720 or 1080 which in essence are then also 0 rotation as a result may break in the same way in Unreal yet we're now not catching those?

Let add it as an issue and move discussion on that later. If this can be exported from maya, but it crashed in unreal. It can be done in either unreal or real. You can clamp the value in Unreal

moonyuet:
> image
> 
> the left most and the right most are wrong. Left one in Unreal is `X:0` `Y:0` `Z:-90` -> `Z` should +90 to look correct. Also the right most in Unreal is `X:0` `Y:0` `Z:90` but the `Z` should be -90.
> 
> So I guess there is still issue of signs.
> 
> Note: in Maya, the left-most is 0, -90, 0 and the right one is 0, +90, 0

Should be working now
discord_image
discord_image_2

antirotor:
Not for me (middle one and the top one):

image
'
I am wondering if we are not running into gimbal lock problem where to rotations "flip". I'll try to run some more test to find out.

moonyuet:
> Not for me (middle one and the top one):
> 
> image ' I am wondering if we are not running into gimbal lock problem where to rotations "flip". I'll try to run some more test to find out.
> 
> The middle one is 0, 0, -180 and the top one -90, 0, -180 in Unreal

Have you pulled the updated commit? (Or I replicate the scene wrong?)

https://github.com/user-attachments/assets/0cec8221-e079-4894-9afb-ded4fcf63844

LiborBatek (changes requested):
There is still some rotation offset on Z axis in UE as seen here

Screenshot 2024-07-30 115746



in maya the spacecraft has different orientation in ROT

Screenshot 2024-07-30 115733

Also worth noting that if I use ```Load Layout as JSON``` it wont bring any assets into UE just uses the existing ones (aka already loaded) ...it used do it automatically before - when some uasset missing it just loaded it in into ```Content / Ayon```

BigRoy:
> There is still some rotation offset on Z axis in UE as seen here

I want to stress again that we should avoid really doing very explicit rotation checks and fixing each 'individual case' as if those are the only problematic areas and we're much better of "finding" the right algorithm that fixes the rotation Maya->Unreal. We referenced an algorithm from Unreal Live Link plug-in somewhere midway through this PR; I still feel that something along those lines is the only real approach... so in essence, finding the right 'math' for the conversion instead of trying to nail it by if statements to resolve very specific edge cases which I assume will (at best) only fix those specific cases and will end up leaving plethora of cases that are still problematic.

moonyuet:
> Also worth noting that if I use `Load Layout as JSON` it wont bring any assets into UE just uses the existing ones (aka already loaded) ...it used do it automatically before - when some uasset missing it just loaded it in into `Content / Ayon`

we can create issues in unreal hosts to fix it.

moonyuet:
@antirotor @LiborBatek can you test this PR along with the unreal PR>

LiborBatek (changes requested):
The layout has still some invalid rotations on asset level. I have tested it with [corresponding PR](https://github.com/ynput/ayon-unreal/pull/35) but that didnt solve it.

image

moonyuet:
> The layout has still some invalid rotations on asset level. I have tested it with [corresponding PR](https://github.com/ynput/ayon-unreal/pull/35) but that didnt solve it.
> 
> image

can you show me the transformation slot to see which rotation has been messed up?

LiborBatek (approved):
This PR works with ```ayon-unreal``` and loading ```Layout``` into ```UE``` results into correct placement / rotation of assets.

image

antirotor (approved):

PR #16, opened 2024-07-08
Collect Look: Include redshift attributes as part of a look

BigRoy:
### Changes

Allow to collect any redshift attributes, prefixed `rs` to be exported along with a Maya Look Publish.

image

### Testing notes

1. Enable Redshift
2. Create a mesh, publish model
3. Load model.
4. Then create the look
5. On the mesh in the lookdev scene, edit any of the Redshift attributes.
6. Publish look.
7. Load the model in new scene,
8. Apply the look.

The attributes on the mesh should now match those set in the look at step 5)


moonyuet (approved):
Tested successfully in Maya

PR #19, opened 2024-07-12
Disallow empty `cbId` value to pass through

BigRoy:
### Issue

If a `cbId` attribute existed on a node, but was an empty value (e.g. user cleared it; I'm looking at you @LiborBatek ) then it would magically pass validation but break in collecting looks.

### Solution

Do not allow empty values to pass through.

### Testing notes

1. Create a model
2. Save scene
3. go to the `cbId` attribute in attribute editor, and clear the value (make it an empty string)
4. Publish the model.

The model should be invalidated and the id be regenerated.

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -1406,7 +1406,12 @@ def get_id_required_nodes(referenced_nodes=False,
     def _add_to_result_if_valid(obj):
         """Add to `result` if the object should be included"""
         fn_dep.setObject(obj)
-        if not existing_ids and fn_dep.hasAttribute("cbId"):
+        if (
+                not existing_ids
+                and fn_dep.hasAttribute("cbId")
+                # May not be an empty value
+                and fn_dep.findPlug("cbId", True).asString()
+        ):
             return
 
         if not referenced_nodes and fn_dep.isFromReferencedFile:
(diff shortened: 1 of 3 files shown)


LiborBatek (changes requested):
It does correctly ```Validate``` if empty ```cbID``` present

Screenshot 2024-08-02 152115

Regenerate action works too but ```Select Invalid``` action does not select anything and inform user in script editor that no entities have been found....

I would like to also propose to change the naming of ```Regenerate UUIDs``` as its refering something else and already caused some confusion to users...but that could be probably separate PR??

```
// pyblish.pyblish.plugin.Action : Finding invalid nodes..
// Warning: pyblish.pyblish.plugin.Action : Plug-in returned to be invalid, but has no selectable nodes.
// pyblish.pyblish.plugin.Action : No invalid nodes found.
```

BigRoy:
> Regenerate action works too but Select Invalid action does not select anything and inform user in script editor that no entities have been found....

Ah, great testing @LiborBatek - this would've actually already been broken BEFORE this PR. Fixed now. ❤️ 

> I would like to also propose to change the naming of Regenerate UUIDs as its refering something else and already caused some confusion to users...but that could be probably separate PR??

Yes, separate issue please - so we can also discuss why and what would be better.

LiborBatek (approved):
Now all works as it should and ```Select Invalid``` too.

Good to go!

image

PR #25, opened 2024-07-16
Create ASS: Add more export options (AY-6100)

BigRoy:
## Changelog Description

Expose more settings to the Create Ass export, like file type specific options for creating compressed `.ass.gz` exports.

## Additional info

New settings
```
ayon+settings://maya/create/CreateAss/maskImager    (default: false)
ayon+settings://maya/create/CreateAss/boundingBox    (default: true)
ayon+settings://maya/create/CreateAss/compressed    (default: false)
```
image

Also on instance:
image

Using the settings as defaults.

### Question: Should I expose "Use Binary Encoding" ❓ 📔 

The Ass file type specific options also expose a "Use Binary Encoding" option:

image

I did not expose it on the Creator because it already seemed tied to a class variable on the Extractor [here](https://github.com/ynput/ayon-maya/blob/472458f8aea2ad15cb71b12c1e6bca08a13d08a2/client/ayon_maya/plugins/publish/extract_arnold_scene_source.py#L15). I wasn't entirely sure whether there was a reason for that - it did not seem to be exposed to settings. Looking at it again it seems that this is because [it's overridden in code for the proxy exports](https://github.com/ynput/ayon-maya/blob/5fd27f373d375ac6871eaec4c172a735c30ff0c5/client/ayon_maya/plugins/publish/extract_arnold_scene_source.py#L167)

@m-u-r-p-h-y Do we want to expose that too?
    - Also toggle-able on the instance? 
    - or should we keep the setting on the Extractor, also in settings hidden from the artists?

#### Improved validator

Also improved a validation report because I had no idea what it did when I hit the error - so I added some description to the report to be clearer for future me and other artists:

image

## Testing notes:

1. Upload new maya addon package
3. Toggle settings should work
4. Export settings should work
5. Resulting publish should work and adhere to the instance settings
6. Loading the published instances should work as intended.

---

Clickup AY-6100


moonyuet (approved):
Done simple test in Maya 2025
With compressed on, it will export .ass.gz
image
When compressed off, it will export .ass
image
Testing with other two new options and working smoothly

m-u-r-p-h-y (commented):
setting the compress format works correctly

my question is why do we have this need to name things differently in every single place?
image

Mask Shape -> Export Shape -> Shape

can't we just match the options in Maya so it is clear what is what?

BigRoy:
> setting the compress format works correctly
> 
> my question is why do we have this need to name things differently in every single place? image
> 
> Mask Shape -> Export Shape -> Shape
> 
> can't we just match the options in Maya so it is clear what is what?

It's because we 'matched' what Arnold does. The actual flags in the arnold export commands name them "mask" whatever. That's where I suspect that originated from, they just don't show that in the labels.

I'll update the settings labels and then we're good?

BigRoy:
> I'll update the settings labels and then we're good?

Done with https://github.com/ynput/ayon-maya/pull/25/commits/252ba09eb1a882eb686faf5d0dec2eb48ab634aa

BigRoy:
@m-u-r-p-h-y can you confirm the latest changes make it work the way you want?

m-u-r-p-h-y (approved):
> @m-u-r-p-h-y can you confirm the latest changes make it work the way you want?

do not like the Export some1, Export some2, Export some3 but we probably can not design the publisher UI in a similar way to Maya. 

the new defaults also match Maya so yes, good to go . . .

BigRoy:
> do not like the Export some1, Export some2, Export some3 but we probably can not design the publisher UI in a similar way to Maya.

@m-u-r-p-h-y only other thing I can get it to is this:

**Alternative**

image

As opposed to:

**Current**

image

---

Would you like me to push the changes so it looks like "Alternative" (first image)?

m-u-r-p-h-y:
> Would you like me to push the changes so it looks like "Alternative" (first image)?

Alternative RULEZ!

PR #27, opened 2024-07-17
Fix bug where maya rig behaves different when loaded due to display handle logic

BigRoy:
## Changelog Description 

- Perform `maya.cmds.refresh` before `maya.cmds.exactWorldBoundingBox` call to resolve issue with rig behaving different on load
- Only perform the logic to set display handle if it's enabled in settings
- Cosmetics/cleanup of code.

## Additional info

For whatever reason the example rig failed to behave normally if we'd call `maya.cmds.exactWorldBoundingBox` directly 

## Testing notes:

1. Maya rig load should work without issue
2. Other reference loads with display handle enabled/disabled should also still work.

This may need a particular "rig" file from Advanced Skeleton to test the _actual issue_ being solved with this PR. However, the file provided for developing this PR was confidential.

### DIFF
--- a/client/ayon_maya/plugins/load/load_reference.py
+++ b/client/ayon_maya/plugins/load/load_reference.py
@@ -212,8 +212,12 @@ class ReferenceLoader(plugin.ReferenceLoader):
                 ).get('display_handle', True)
                 if display_handle:
                     cmds.setAttr(f"{group_name}.displayHandle", display_handle)
-                    
                     # get bounding box
+                    # Bugfix: We force a refresh here because there is a
+                    # reproducable case with Advanced Skeleton rig where the
+                    # call to `exactWorldBoundingBox` directly after the
+                    # reference without it breaks the behavior of the rigs
+                    # making it appear as if parts of the mesh are static.
                     cmds.refresh()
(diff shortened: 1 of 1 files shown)


BigRoy:
Example rig producing the issue:

[9989_test_ayon_pipeline_asset001_rigMain_v003.zip](https://github.com/user-attachments/files/16291509/9989_test_ayon_pipeline_asset001_rigMain_v003.zip)
image

And to simply reproduce the issue OUTSIDE of the pipeline:

```python
from maya import cmds

file_url = r"/path/to/9989_test_ayon_pipeline_asset001_rigMain_v003.ma"
nodes = cmds.file(file_url,
                  namespace="test",
                  sharedReferenceFile=False,
                  reference=True,
                  returnNewNodes=True,
                  groupReference=True,
                  groupName="my_group")
cmds.exactWorldBoundingBox(nodes)  # this breaks the rig directly after the `file` call
```

m-u-r-p-h-y (approved):
tested by the client and confirmed it solves the issue

BigRoy:
I've reported this as a bug to autodesk as well.

> Case number
> 22606091

PR #33, opened 2024-07-18
Connect Geometry: Do not error on non-unique node names

BigRoy:
Fixes an issue using Inventory action "Connect Geometry" where if the hierarchy of the loaded containers contains non-unique node names it would result in an error.

Original error:
```python
#   File "C:\Users\USER\AppData\Local\Ynput\AYON\addons\openpype_3.18.7\openpype\tools\ayon_sceneinventory\view.py", line 458, in _process_custom_action
#     result = action.process(containers)
#   File "C:\Users\USER\AppData\Local\Ynput\AYON\addons\openpype_3.18.7\openpype\hosts\maya\plugins\inventory\connect_geometry.py", line 66, in process
#     source_data = self.get_container_data(source_object)
#   File "C:\Users\USER\AppData\Local\Ynput\AYON\addons\openpype_3.18.7\openpype\hosts\maya\plugins\inventory\connect_geometry.py", line 117, in get_container_data
#     node_type = cmds.nodeType(node)
# RuntimeError: nodeType: No object or UFE path string matches name 'NAMESPACE:_GRP'.
```

Example hierarchy:
example

### DIFF
--- a/client/ayon_maya/plugins/inventory/connect_geometry.py
+++ b/client/ayon_maya/plugins/inventory/connect_geometry.py
@@ -1,7 +1,7 @@
 from maya import cmds
 
 from ayon_core.pipeline import InventoryAction, get_repres_contexts
-from ayon_maya.api.lib import get_id
+from ayon_maya.api.lib import get_id, get_container_members
 
 
 class ConnectGeometry(InventoryAction):
@@ -117,8 +117,7 @@ class ConnectGeometry(InventoryAction):
                     id.
         """
         data = {"node_types": set(), "ids": {}}
-        ref_node = cmds.sets(container, query=True, nodesOnly=True)[0]
-        for node in cmds.referenceQuery(ref_node, nodes=True):
+        for node in get_container_members(container):
             node_type = cmds.nodeType(node)
             data["node_types"].add(node_type)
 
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Took me a while  to prepare such conditions (aka rig with the same names inside the hierarchy etc) but at the end I have succesfully failed first :) without this branch as seen below:

Screenshot 2024-07-30 164817

when using this branch all working well and no errors shown and geometry got connection with Xgen rig...

Screenshot 2024-07-30 165534

So I guess good to go!

LiborBatek:
@BigRoy just a question, do u think that restriction to be able to connect xgen just with ```animation``` and/or ```pointcache``` has real reason? as I can imagine it could be quite useful to be able to connect it with actual ```rig``` asset too....as now we do not support such ```action```

BigRoy:
> just a question, do u think that restriction to be able to connect xgen just with `animation` and/or `pointcache` has real reason? as I can imagine it could be quite useful to be able to connect it with actual `rig` asset too....as now we do not support such `action`

I agree - the restrictions may be a bit too strict. Let's track it as low priority issue; just so we have it for whenever a client requests it or needs it. For now it hasn't seem to have gotten in the way of production for anyone yet.

PR #37, opened 2024-07-22
Support to create face sets for materials in Extract Alembic

moonyuet:
### Changelog Descriptions
This PR is to add support for add materials per face sets for in Extract Alembic when the "write face sets" option is enabled. It also adds the optional validator to check on the write face sets options. Users can turn on the validator when they want to export the alembic to Unreal Engine to make sure the write face sets options switched on.

If the `write face sets` option is enabled with this PR, it will assign the material slots per each mesh while the mesh is merged.
image
If the `write face sets` option is disabled, it will result one shader in merged mesh
image

### Additional Info
It is an extension of fix for this issue lasted in ayon-unreal
https://github.com/ynput/ayon-unreal/issues/21#issuecomment-2242540494
The related enhancement of the PR created for ayon_unreal: https://github.com/ynput/ayon-unreal/pull/25
### Testing Notes
1. Launch Maya
2. Create Model/Point Cache
3. Make sure the write face sets turned on
4. Publish
5. Loaded the model in Unreal
6. There should be multiple material slots if there are materials/shaders assigned.


client/ayon_maya/api/lib.py
```diff
@@ -4343,3 +4344,69 @@ def get_sequence(filepath, pattern="%04d"):
...
+
+
+@contextlib.contextmanager
+def write_face_sets_for_materials(shapes):
+    all_write_face_sets = create_face_sets_for_materials(shapes)
+    try:
+        yield
+    finally:
+        for write_face_sets in all_write_face_sets:
+            cmds.delete(write_face_sets)
```
BigRoy, line 4356:
```suggestion
        cmds.delete(all_write_face_sets)
```

client/ayon_maya/plugins/publish/extract_model.py
```diff
@@ -76,9 +77,12 @@ def process(self, instance):
...
                                        polygonObject=1):
-                with lib.shader(members,
-                                shadingEngine="initialShadingGroup"):
-                    with lib.maintained_selection():
+                with lib.maintained_selection():
+                    with contextlib.ExitStack() as stack:
+                        if instance.data.get("writeFaceSets", True):
+                            stack.enter_context(
+                                lib.write_face_sets_for_materials(members)
+                            )
```
BigRoy, line 85: If we are using a `ExitStack` now I think it's a good time to just move all the context managers into the `ExitStack`.
That would also avoid the deep nesting we have now.

client/ayon_maya/plugins/publish/validate_no_face_sets.py
```diff
@@ -0,0 +1,50 @@
...
+            return
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise PublishValidationError(
+                message="Write Face Sets Disabled",
+                title="No Write Face Sets found in mesh",
+                description=(
+                    "## Write Face Sets Disabled\n"
+                    "The mesh(es) require to have face sets to "
+                    "generate material with multiple shading entity."
```
BigRoy, line 48: This does not explain WHY it is needed and what issues might occur otherwise.

Preferably these validations explain what is going on and the reasoning on why it needs fixing.

client/ayon_maya/plugins/publish/validate_no_face_sets.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise PublishValidationError(
+                message="Write Face Sets Disabled",
+                title="No Write Face Sets found in mesh",
```
BigRoy, line 44: This title seems weird since the write face sets is not set in the mesh.

- "Write face sets is disabled"
- "Multiple materials on meshes"

client/ayon_maya/plugins/publish/validate_no_face_sets.py
```diff
@@ -0,0 +1,50 @@
...
+    OptionalPyblishPluginMixin,
+    PublishValidationError,
+    RepairAction,
+    ValidateContentsOrder,
+)
+from ayon_maya.api import plugin
+
+class ValidateNoFaceSets(plugin.MayaInstancePlugin,
+                         OptionalPyblishPluginMixin):
+    """Ensure the nodes don't have a namespace"""
```
BigRoy, line 13: Docstring incorrect.

client/ayon_maya/plugins/publish/validate_no_face_sets.py
moonyuet: Agreed. I forgot to mention unreal engine import needed these maps.

client/ayon_maya/plugins/publish/validate_no_face_sets.py
```diff
@@ -0,0 +1,51 @@
...
+        invalid = []
+        members = instance.data["setMembers"]
+        members = cmds.ls(members,
+                          dag=True,
+                          shapes=True,
+                          type=("mesh", "nurbsCurve"),
+                          noIntermediate=True,
+                          long=True)
+        if not instance.data.get("writeFaceSets"):
+            invalid.append(members)
```
BigRoy, line 33: This validation is incorrect - it does not check at all whether the meshes have component shader assignments. This will now always invalidate if there are any meshes (or curves!) which to me seems wrong.

1. First, curves can have no face/component shading assignments, right? (or even any shader assignment in maya?) so should not need to be checked.
2. We should only find the meshes that have face shader assignments to begin with?

Otherwise there's no reason to make this a validator and instead force the `writeFaceSets` to True always.

BigRoy: For reference, here's a similar validator for maya looks: https://github.com/ynput/ayon-maya/blob/06aa2445b4df511225b6795b283dd974040c0b5f/client/ayon_maya/plugins/publish/validate_look_single_shader.py#L1-L57

Basically disallowing shaders per face to be assigned. You can perform a similar check here and consider only the meshes with multiple shaders to have face assignments.

client/ayon_maya/plugins/publish/validate_no_face_sets.py
```diff
@@ -0,0 +1,51 @@
...
+    OptionalPyblishPluginMixin,
+    PublishValidationError,
+    RepairAction,
+    ValidateContentsOrder,
+)
+from ayon_maya.api import plugin
+
+class ValidateNoFaceSets(plugin.MayaInstancePlugin,
+                         OptionalPyblishPluginMixin):
+    """Ensure the meshes don't have face sets """
```
BigRoy, line 13: I have my doubts about the name and docstring. It seems to solely be about "Shader assignments per face" - not about "Face Sets" so I think we're better off wording this in a better way that it refers to shader assignments per face or alike.

Or at least to me it's confusing. I'd never heard of "Face Sets" in Maya before. :D

moonyuet: It would be more concise to call "shaders per face". But the alembic options called "write face sets"

BigRoy (commented):
You have now completely removed the validator and tried to force it to generate the face sets - but it doesn't - it just "defaults" to it enabled as such an artist could still break it intentionally. (And existing instances will still have it off because they already had the attribute.)

I don't use the feature myself, but is that now sufficient? Also, the logic does NOT create face sets for objects that do not have face assignments (is that the correct intended behavior as well?; I wrote that code but admittedly I have no clue what it's for in the end)

moonyuet:
> You have now completely removed the validator and tried to force it to generate the face sets - but it doesn't - it just "defaults" to it enabled as such an artist could still break it intentionally. (And existing instances will still have it off because they already had the attribute.)
> 
> I don't use the feature myself, but is that now sufficient? Also, the logic does NOT create face sets for objects that do not have face assignments (is that the correct intended behavior as well?; I wrote that code but admittedly I have no clue what it's for in the end)

It is for creating face sets for the alembic export. The current issue is when we import the static mesh in alembic format to Unreal Engine, they can't find the face sets for the material which give the warnings. If the users import the mesh without the face sets, they can't assign materials correctly in the merged mesh sets as there is only single material slot provided in the objects in UE(If there are face sets found in the object, there should be multiple material slots). 

My initial thought is to use the validator to check if there are face sets for the materials of the set members and if there isn't, we can use the repair action to assign/create the face sets so that all the face sets can be loaded correctly inside unreal engine. 
I mentioned the same [link](https://www.reddit.com/r/Maya/comments/om2v70/script_to_create_face_sets_from_all_materials/) which is a good reference for this PR in the related issue https://github.com/ynput/ayon-unreal/issues/21#issuecomment-2242540494,

client/ayon_maya/plugins/publish/extract_pointcache.py
```diff
@@ -260,6 +267,16 @@ def process(self, instance):
...
 
+    def get_members_shapes(self, instance):
+        members = instance.data("setMembers")
+        members = cmds.ls(members,
+                          dag=True,
+                          shapes=True,
+                          type=("mesh", "nurbsCurve"),
+                          noIntermediate=True,
+                          long=True)
+        return members
```
BigRoy, line 278: FYI, This logic may fail for instanced geometry. Also shouldn't the members already have been collected by the collector, essentially making this:

```
nodes = list(instance[:])
return cmds.ls(nodes, type=("mesh", "nurbsCurve"), long=True)
```
Right?

client/ayon_maya/api/lib.py
```diff
@@ -4343,3 +4343,39 @@ def get_sequence(filepath, pattern="%04d"):
...
+            #store a list of the selected objects
+            object_with_mat = cmds.ls(selection=True)
+            #assign lambert1 (a material that should always exist) to selection
+            cmds.hyperShade(assign="lambert1")
+            #clear selection
+            cmds.select(deselect=True)
+
+            for shape in object_with_mat:
+                cmds.select(shape +".f[*]",add=True)
+            cmds.hyperShade(assign=shading_engine)
```
BigRoy, line 4373: Sorry - wait a second. What are we doing here, and why? What!?

1. Lambert1 is not required to exist, especially now in recent Maya versions where it defaults to a standard surface.
2. Why does this use a selection? What?
3. Do not use the `hypershade` command for what is essentially an assignment of a set.

What was wrong with the previous implementation where it generates object sets?

client/ayon_maya/api/lib.py
```diff
@@ -4343,3 +4343,39 @@ def get_sequence(filepath, pattern="%04d"):
...
+    Activated only when `writeFaceSets` alembic options enabled.
+
+    Args:
+        shapes (List[str]): The shapes to add into face sets for any component
+            assignments of shading engines.
+    """
+    original_material_list = {}
+    for shape in shapes:
+        shading_engines = cmds.listConnections(shape, destination=True, type="shadingEngine") or []
+        original_material_list.update({shape: shading_engines})
```
BigRoy, line 4360: Wait - you're breaking something here. If a shape previously had mutliple shaders we need to known WHICH components had which shader - not just reapply the shader to the full mesh afterwards.

You're breaking something here.

BigRoy (changes requested):

client/ayon_maya/api/lib.py
```diff
@@ -4343,3 +4343,39 @@ def get_sequence(filepath, pattern="%04d"):
...
+                cmds.select(shape +".f[*]",add=True)
+            cmds.hyperShade(assign=shading_engine)
+    try:
+        yield
+
+    finally:
+            for original_shape, shading_engines in original_material_list.items():
+                cmds.select(original_shape)
+                for shading_engine in shading_engines:
+                    cmds.hyperShade(assign=shading_engine)
```
BigRoy, line 4381:
```python


@contextlib.contextmanager
def force_shader_assignments_to_faces(shapes):
    
    shapes = cmds.ls(shapes, shapes=True, type="mesh", long=True)
    if not shapes:
       # Do nothing
       yield
       return
    
    all_render_sets = set()
    for shape in shapes:
        render_sets = cmds.listSets(object=shape, t=1, extendToShape=False)
        if render_sets:
            all_render_sets.update(render_sets)
            
    shapes_lookup = set(shapes)
    original_assignments = {}
    override_assignments = defaultdict(list)
    for shading_engine in cmds.ls(list(all_render_sets), type="shadingEngine"):
        members = cmds.sets(shading_engine, query=True)
        if not members:
            continue
            
        members = cmds.ls(members, long=True)
        
        # Include ALL originals, even those not among our shapes
        original_assignments[shading_engine] = members
        
        for member in members:
            
            # Ignore component assignments because they are already the
            # the way we want
            if "." in member:
                continue
                
            # Only consider shapes from out inputs
            if member not in shapes_lookup:
                continue
                
            member_faces = f"{member}.f[*]"
            override_assignments[shading_engine].append(member_faces)
            
    try:
        # Apply overrides
        for shading_engine, members in override_assignments.items():
            cmds.sets(clear=shading_engine)
            cmds.sets(members, forceElement=shading_engine)
            
        yield
        
    finally:
        # Revert to original assignments
        for shading_engine, members in original_assignments.items():
            cmds.sets(clear=shading_engine)
            cmds.sets(members, forceElement=shading_engine)
```

This should be closer to what you're looking for @moonyuet ?

moonyuet: thank you! it works ❤️

client/ayon_maya/api/lib.py
```diff
@@ -4343,3 +4343,61 @@ def get_sequence(filepath, pattern="%04d"):
...
+            cmds.sets(clear=shading_engine)
+            cmds.sets(members, forceElement=shading_engine)
+
+        yield
+
+    finally:
+        # Revert to original assignments
+        for shading_engine, members in original_assignments.items():
+            cmds.sets(clear=shading_engine)
+            cmds.sets(members, forceElement=shading_engine)
```
BigRoy, line 4403: So, looking at this once again - I believe this is actually buggy. Because below we're clearing the set but it seems Maya doesn't do that clearing for face assignments. Overall, this is NOT problematic... but it does become problematic as soon as Maya ever fixes that bug :)

The issue we're having is that we're throwing out all members then which are already face assignments. (which would only be problematic for those shaders that have both entries with components and on other meshes components without because it'd clear for that set but forget to add the entries with the existing component assignments.

```suggestion@contextlib.contextmanager
def force_shader_assignments_to_faces(shapes):
    """Replaces any non-face shader assignments with shader assignments
    to the faces during the context.

    Args:
        shapes (List[str]): The shapes to add into face sets for any component
            assignments of shading engine
    """
    shapes = cmds.ls(shapes, shapes=True, type="mesh", long=True)
    if not shapes:
        # Do nothing
        yield
        return

    all_render_sets = set()
    for shape in shapes:
        render_sets = cmds.listSets(object=shape, t=1, extendToShape=False)
        if render_sets:
            all_render_sets.update(render_sets)

    shapes_lookup = set(shapes)

    # Maya has the tendency to return component assignment using the transform
    # name instead of the shape name, like `pCube1.f[1]` instead of
    # `pCube1Shape.f[1]` so we need to take those into consideration as members
    def get_parent(_shape: str) -> str:
        return _shape.rsplit("|", 1)[0]

    components_lookup = {f"{shape}." for shape in shapes}
    components_lookup.update(f"{get_parent(shape)}." for shape in shapes)
    components_lookup = tuple(components_lookup)  # support str.startswith

    original_assignments = {}
    override_assignments = defaultdict(list)
    for shading_engine in cmds.ls(list(all_render_sets), type="shadingEngine"):
        members = cmds.sets(shading_engine, query=True)
        if not members:
            continue

        members = cmds.ls(members, long=True)

        # Include ALL originals, even those not among our shapes
        original_assignments[shading_engine] = members

        has_conversions = False
        for member in members:
            # Only consider shapes from our inputs
            if (
                    member not in shapes_lookup
                    and not member.startswith(components_lookup)
            ):
                continue

            if "." not in member:
                # Convert to face assignments
                member = f"{member}.f[*]"
                has_conversions = True
            override_assignments[shading_engine].append(member)

        if not has_conversions:
            # We can skip this shading engine completely because
            # we have nothing to override
            original_assignments.pop(shading_engine, None)
            override_assignments.pop(shading_engine, None)

    try:
        # Apply overrides
        for shading_engine, override_members in override_assignments.items():
            # We force remove the members because this allows maya to take
            # out the mesh (also without the components)
            cmds.sets(clear=shading_engine)
            cmds.sets(override_members, forceElement=shading_engine)

        yield

    finally:
        # Revert to original assignments
        for shading_engine, original_members in original_assignments.items():
            cmds.sets(clear=shading_engine)
            cmds.sets(original_members, forceElement=shading_engine)
```

PR #38, opened 2024-07-23
Bugfix: Update instance frame ranges on context change

BigRoy:
## Changelog Description

Fix updating of frame ranges for the publish instances when saving the maya scene into another context and pressing OK with "Publish instances" checkbox enabled in the update to new context dialog.

image

## Additional info

- It was basically a typo where it was trying to set the value into the `instance` instead of the `creator_attributes`.
- When fixed, I also noticed the attributes turned into integers even though by default they are floats - so I converted the values explicitly.
- The variable name `folder_entity` was used for what was the `task_entity` so renamed that as well.

### Question 1: Should it also update frame ranges?

The code does not update the handle start or handle end. And I wonder, should it? If so, we'd need to add:
```python
        "handleStart": float(task_entity["attrib"]["handleStart"]),
        "handleEnd": float(task_entity["attrib"]["handleEnd"]),
```

### Question 2: Should updating of the frame ranges become a separate checkbox from updating the target context (folder/task) in the instance?

Would there be cases where you do want to update the instance to the new folder+task but do not want to update the frame ranges for the instances to the new context? If so, should it be a separate checkbox in the dialog?

## Testing notes:

1. Create a scene with instances that have frame ranges (e.g. review, pointcache, animation instances)
2. Save into another context with a different frame range on the task entity
3. Pressing ok on the 'update to changed context prompt' should update the frame ranges on the publish instances.

---

Clickup AY-6148

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -3199,15 +3199,15 @@ def update_content_on_context_change():
 
     host = registered_host()
     create_context = CreateContext(host)
-    folder_entity = get_current_task_entity(fields={"attrib"})
+    task_entity = get_current_task_entity(fields={"attrib"})
 
     instance_values = {
         "folderPath": create_context.get_current_folder_path(),
         "task": create_context.get_current_task_name(),
     }
     creator_attribute_values = {
-        "frameStart": folder_entity["attrib"]["frameStart"],
-        "frameEnd": folder_entity["attrib"]["frameEnd"],
+        "frameStart": task_entity["attrib"]["frameStart"],
+        "frameEnd": task_entity["attrib"]["frameEnd"],
     }
 
     has_changes = False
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (approved):
frame end was correctly applied when shot010 was saved to sh020 context and task override was used

image

moonyuet (approved):

BigRoy:
> ### Question 1: Should it also update frame ranges?
> 
> The code does not update the handle start or handle end. And I wonder, should it? If so, we'd need to add:
> 
> ```python
>         "handleStart": float(task_entity["attrib"]["handleStart"]),
>         "handleEnd": float(task_entity["attrib"]["handleEnd"]),
> ```
> 
> ### Question 2: Should updating of the frame ranges become a separate checkbox from updating the target context (folder/task) in the instance?
> 
> Would there be cases where you do want to update the instance to the new folder+task but do not want to update the frame ranges for the instances to the new context? If so, should it be a separate checkbox in the dialog?

@m-u-r-p-h-y could you still check in regarding these questions? potentially for a follow up PR?

PR #40, opened 2024-07-26
Pointcache/Animation: Adhere to the "visible only" toggle

BigRoy:
## Changelog Description

Actually trigger the visible only logic when it is enabled.

## Additional info

With Visible Only enabled for the alembic exports it should ONLY include objects that are visible at least once in the timeline during the to be exproted frame range.

So create a group with four cubes:

## Testing notes:

1. Create some cubes with visibilities.
```
grp/
    cube_visible
    cube_invisible
    cube_visible_in_range
    cube_visible_outside_range
```

2. And then publish the frame range with visible only enabled.
3. Load back in. Only the cubes visible inside the frame range should be included in the export when loading back in.

---

Clickup AY-6188

### DIFF
--- a/client/ayon_maya/plugins/publish/collect_pointcache_visible_only.py
+++ b/client/ayon_maya/plugins/publish/collect_pointcache_visible_only.py
@@ -1,0 +1,26 @@
+import pyblish.api
+from ayon_maya.api import plugin
+
+
+class CollectPointcacheVisibleOnly(plugin.MayaInstancePlugin):
+    """Collect pointcache visible only data for instance."""
+
+    order = pyblish.api.CollectorOrder + 0.4
+    families = ["pointcache", "animation", "model", "vrayproxy.alembic"]
+    label = "Collect Pointcache Visible Only"
+
+    def process(self, instance):
+
+        if instance.data["productType"] == "animation":
+            plugin_name = "ExtractAnimation"
+        else:
+            plugin_name = "ExtractAlembic"
+
+        publish_attributes = instance.data.get("publish_attributes", {})
+        plugin_attr_values = publish_attributes.get(plugin_name, {})
+
+        if "visibleOnly" in plugin_attr_values:
+            value = plugin_attr_values["visibleOnly"]
+            instance.data["visibleOnly"] = value
+            self.log.debug(
+                f"Transferring visibleOnly to instance.data: {value}")
--- a/client/ayon_maya/plugins/publish/extract_pointcache.py
+++ b/client/ayon_maya/plugins/publish/extract_pointcache.py
@@ -195,7 +195,7 @@ class ExtractAlembic(plugin.MayaExtractorPlugin, AYONPyblishPluginMixin):
             ),
         }
 
-        if instance.data.get("visibleOnly", False):
+        if attribute_values.get("visibleOnly", False):
             # If we only want to include nodes that are visible in the frame
             # range then we need to do our own check. Alembic's `visibleOnly`
             # flag does not filter out those that are only hidden on some
--- a/client/ayon_maya/plugins/publish/validate_alembic_options_defaults.py
+++ b/client/ayon_maya/plugins/publish/validate_alembic_options_defaults.py
@@ -40,10 +40,6 @@ class ValidateAlembicDefaultsPointcache(
         settings = self._get_settings(instance.context)
         attributes = self._get_publish_attributes(instance)
 
-        import pprint
-        self.log.info(pprint.pformat(settings))
-        self.log.info(pprint.pformat(attributes))
-
         invalid = {}
         for key, value in attributes.items():
             if key not in settings:
--- a/client/ayon_maya/plugins/publish/validate_alembic_options_defaults.py
+++ b/client/ayon_maya/plugins/publish/validate_alembic_options_defaults.py
@@ -3,6 +3,7 @@ import inspect
 import pyblish.api
 from ayon_core.pipeline import OptionalPyblishPluginMixin
 from ayon_core.pipeline.publish import PublishValidationError, RepairAction
+from ayon_core.pipeline.create.context import PublishAttributeValues
 from ayon_maya.api import plugin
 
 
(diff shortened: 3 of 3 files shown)


BigRoy:
I'm starting to think this is the wrong fix because this plug-in also expects it in `instance.data` https://github.com/ynput/ayon-maya/blob/06aa2445b4df511225b6795b283dd974040c0b5f/client/ayon_maya/plugins/publish/validate_visible_only.py#L28

BigRoy:
Unfortunately had to opt to doing this https://github.com/ynput/ayon-maya/pull/40/commits/78e598486ab8aaf8c5e5cef712c398d04724d6c6 instead.

BigRoy:
Works on my end. And one of Ynput clients reports this PR does indeed fix the issue that `visibleOnly` actually does something.

moonyuet:
Testing in my end, it seems the cube_visible_in_range has some issues though.
https://github.com/user-attachments/assets/065e60fb-d329-4048-8608-b2097ac4fed3

Also, it errors out when validate alembic options default tries to repair visibleOnly options.
image

BigRoy:
> Testing in my end, it seems the cube_visible_in_range has some issues though.
> https://github.com/user-attachments/assets/065e60fb-d329-4048-8608-b2097ac4fed3

So - in this case, I'm pretty sure you are writing the Alembics without **Write Visibility** enabled.

image

```
ayon+settings://maya/publish/ExtractAlembic/writeVisibility
```

I have no idea why these are disabled by default (it's not really a sensible default for production, but it does match Maya's export alembic export dialog settings defaults)

---

> Also, it errors out when validate alembic options default tries to repair visibleOnly options.

That shouldn't be due to this PR actually... 
If you hit these types of errors, please include the stack trace to make it easier to debug what's going on. (So in this case sharing the full details of that error) - preferably also as code block instead of just an image to allow searching for it in the future as well from Github - which may help anyone else hitting the issue down the line as well.

I can confirm the error here:
```python
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\dev\ayon-core\client\ayon_core\pipeline\publish\publish_plugins.py", line 243, in process
    plugin.repair(instance)
  File "E:\dev\ayon-maya\client\ayon_maya\plugins\publish\validate_alembic_options_defaults.py", line 105, in repair
    for key in attributes:
  File "E:\dev\ayon-core\client\ayon_core\pipeline\create\context.py", line 576, in __getitem__
    return self._data[key]
KeyError: 0
```
Should be fixed with: https://github.com/ynput/ayon-maya/pull/40/commits/35aab17d62cb2abac81a532e6c63a7852d398928

moonyuet:
@BigRoy there is still some issues on writing visibilities. (I have enabled both writing visibilities and visibilities only)
https://github.com/user-attachments/assets/40a59e77-6b73-4e1e-89a8-8b039fa9508d

The option I used for exporting the alembic manually.
image

BigRoy:
> @BigRoy there is still some issues on writing visibilities. (I have enabled both writing visibilities and visibilities only) https://github.com/user-attachments/assets/40a59e77-6b73-4e1e-89a8-8b039fa9508d
> 
> The option I used for exporting the alembic manually. image
> 
> But the repair action has been fixed.

@moonyuet Could you show me where you enabled it in settings and what the value is?
And/or share a JSON log of your publish to confirm the settings do as intended? It works for me (and for a client) so I have a feeling this must be a configuration error instead.

moonyuet:
> > @BigRoy there is still some issues on writing visibilities. (I have enabled both writing visibilities and visibilities only) https://github.com/user-attachments/assets/40a59e77-6b73-4e1e-89a8-8b039fa9508d
> > The option I used for exporting the alembic manually. image
> > But the repair action has been fixed.
> 
> @moonyuet Could you show me where you enabled it in settings and what the value is? And/or share a JSON log of your publish to confirm the settings do as intended? It works for me (and for a client) so I have a feeling this must be a configuration error instead.

`ayon+settings://maya/publish/ExtractAlembic/writeVisibility`
image
[publish-report-240729-20-51.json](https://github.com/user-attachments/files/16413303/publish-report-240729-20-51.json)

BigRoy:
> The report: [publish-report-240729-20-51.json](https://github.com/user-attachments/files/16413303/publish-report-240729-20-51.json)

So the log mentions it does write the visibility:

```
DEBUG: Running `extract_alembic` with the keyword arguments: {'file': 'C:\\Users\\Kayla\\AppData\\Local\\Temp\\pyblish_tmp_ezv85sfd\\pointcacheMain.abc', 'attr': ['cbId'], 'attrPrefix': [], 'userAttr': [], 'userAttrPrefix': [], 'dataFormat': 'ogawa', 'endFrame': 1001.0, 'eulerFilter': False, 'preRoll': False, 'preRollStartFrame': 0, 'renderableOnly': False, 'root': ['|cube_GRP|cube_visible_in_range_GEO', '|cube_GRP|cube_visible_outside_range_GEO', '|cube_GRP|cube_invisible_GEO', '|cube_GRP|cube_visible_GEO', 'pointcacheMain_proxy_SET'], 'selection': True, 'startFrame': 1001.0, 'step': 1.0, 'stripNamespaces': True, 'uvWrite': True, 'verbose': False, 'wholeFrameGeo': False, 'worldSpace': True, 'writeColorSets': False, 'writeCreases': False, 'writeFaceSets': False, 'writeUVSets': False, 'writeVisibility': True, 'uvsOnly': False, 'melPerFrameCallback': '', 'melPostJobCallback': '', 'pythonPerFrameCallback': '', 'pythonPostJobCallback': '', 'noNormals': False}
```
So that's good.

It also correctly triggers the `visibleOnly` argument.
```
DEBUG: Transferring visibleOnly to instance.data: True
```

And that's also why you get just the two cubes ✅ 

The question is now, why doesn't your publish have keys?

1. Could it be that BOTH your cubes are visible (and never invisible) during the FULL frame range? And as such, they get no keys because the visibility never 'changes' within the frame range?

So that will be 100% the case ☝️ because you are exporting 1001-1001 so if it would ever be visible, it would always be visible during that "full frame range" and hence, would get no visibility key.

As such, it works as intended. 👍

PR #59, opened 2024-07-30
Improve Maya USD Extraction

BigRoy:
## Changelog Description

Add more options to the Maya USD Extractor:
- Expose USD extractor for model and pointcache family
- Do not export shaders by default
- Allow to Enable/disable animation data
- Update defaults to be more sensible like maya's defaults
- Allow using export roots to exclude parents in export
- Allow to export in worldspace when parents are not included
- Ignore constraints to avoid constraints breaking merge transform/shape behavior

## Additional info

Separated from #2

## Testing notes:

1. Maya USD extraction should work
2. When enabled, model extraction with Extract `usd` enabled should work
3. When enabled, pointcache extraction with Extract `usd` enabled should work
4. When enabled, animation extraction with Extract `usd` enabled should work
5. The Maya USD Creator has a **Create Asset Hierarchy** checkbox (on Create tab) which should do what it describes:

image

### DIFF
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -2,21 +2,22 @@ from ayon_maya.api import plugin, lib
 from ayon_core.lib import (
     BoolDef,
     EnumDef,
-    TextDef
+    TextDef,
+    UILabelDef,
+    UISeparatorDef,
 )
 
 from maya import cmds
 
 
 class CreateMayaUsd(plugin.MayaCreator):
-    """Create Maya USD Export"""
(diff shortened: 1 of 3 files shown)


moonyuet (commented):
Tested with animation product type
looks good and animation publish. And I can load back the animation data as expected.
Will test the model/pointcache product type later

moonyuet (approved):
Added screenshot regarding to the previous comment
image

Tested in Model and Pointcache product type, published and loaded without any error. 
also tested with the maya usd product type publish and load. looks good.
Animation, pointcache and usd can export animation data as expected when export animation data being turned on. 

image

PR #60, opened 2024-07-30
Implement Maya USD Proxy extractor for one of the layers in the proxy

BigRoy:
## Changelog Description

Add Creator to extract a single USD layer from a Maya USD Proxy Shape as output data.

## Additional info

This would be used to do e.g. an export of LookdevX (Maya's USD shader assignments) materials with their bindings to the geometry.

Separated from #2 

## Testing notes:

1. Creator should work
2. Instance should allow you to pick _which_ USD maya proxy shape layer from the scene to export

### DIFF
--- a/client/ayon_maya/plugins/create/create_maya_usd_layer.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd_layer.py
@@ -1,0 +1,59 @@
+from ayon_maya.api import plugin
+from ayon_core.lib import EnumDef
+
+
+class CreateMayaUsdLayer(plugin.MayaCreator):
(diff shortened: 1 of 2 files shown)


moonyuet:
@BigRoy I can't find the the maya usd proxy creator in publisher UI, even though i have the USD layer shape
image

BigRoy:
@moonyuet should be fixed

moonyuet (approved):
Tested in Maya 2024.
In the Publisher UI, we can choose which stage layers we want to export.
image
When only export the selected layers, it will export the layers as expected.

https://github.com/user-attachments/assets/4ab95e30-118a-4695-8581-66be5b2d9234

PR #62, opened 2024-07-31
Extract Maya Scene Raw loaded container would check on NoneType value.

moonyuet:
## Changelog Description
This PR is to fix the publishing error occurs when we load the meshes and publish them as part of the layout product type.
The error shown below:
image

## Additional info
n/a

## Testing notes:

1. Launch Maya
2. Load assets with model product type
3. Create Layout with the loaded asset
4. Publish

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
+++ b/client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
@@ -145,7 +145,7 @@ class ExtractMayaSceneRaw(plugin.MayaExtractorPlugin, AYONPyblishPluginMixin):
                 continue
 
             set_content = set(cmds.sets(obj_set, query=True))
-            if set_content.intersection(members_with_refs):
+            if set_content and set_content.intersection(members_with_refs):
                 loaded_containers.append(obj_set)
 
         return loaded_containers
--- a/client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
+++ b/client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
@@ -3,8 +3,9 @@
 import os
 import contextlib
 from ayon_core.lib import BoolDef
+from ayon_core.pipeline import AVALON_CONTAINER_ID, AYON_CONTAINER_ID
 from ayon_core.pipeline.publish import AYONPyblishPluginMixin
-from ayon_maya.api.lib import maintained_selection, shader, get_container_members
+from ayon_maya.api.lib import maintained_selection, shader
 from ayon_maya.api import plugin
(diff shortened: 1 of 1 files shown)


client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -145,7 +145,7 @@ def _get_loaded_containers(members):
                 continue
 
             set_content = set(cmds.sets(obj_set, query=True))
```
BigRoy, line 147:
```suggestion
            set_content = set(cmds.sets(obj_set, query=True) or [])
```

client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -145,7 +145,7 @@ def _get_loaded_containers(members):
...
-            if set_content.intersection(members_with_refs):
+            if set_content and set_content.intersection(members_with_refs):
```
BigRoy, line 148:
```suggestion
            if set_content.intersection(members_with_refs):
```

BigRoy (changes requested):
You code doesn't fix the error because the error is on the unchanged line above. I've proposed a fix.

BigRoy (approved):
Didn't test but change makes sense. :)

BigRoy:
Not needed for this "hotfix" but this should preferably use the `lib.get_container_members` function call?
AND also, this may be much better if it would also use the actual `host.get_containers()` method to ensure it actually collects containers that the Maya integration itself also defines are containers instead of the custom logic here.

moonyuet:
> Not needed for this "hotfix" but this should preferably use the `lib.get_container_members` function call? AND also, this may be much better if it would also use the actual `host.get_containers()` method to ensure it actually collects containers that the Maya integration itself also defines are containers instead of the custom logic here.

maybe we should use the one from library instead. Much safer and make sense

client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -79,7 +78,7 @@ def process(self, instance):
         if set(self.add_for_families).intersection(
                 set(instance.data.get("families", []))) or \
                 instance.data.get("productType") in self.add_for_families:
-            selection += self._get_loaded_containers(members)
+            selection += get_container_members(members)
```
BigRoy, line 81: This change is wrong.. we still need to first get the containers that are members of this instance... and then go from there. The logic you have here now will error.

BigRoy (approved):
Looks good to me.

PR #69, opened 2024-08-16
feature: add option for including shaders on model export.

nikshankarnoble:
## Changelog Description

This is a duplication of the old PR https://github.com/ynput/ayon-core/pull/444 which got closed as part of the migration to the new repository. 

The current default behaviour when exporting a model is that the shaders from Maya will be stripped, but for certain workflows (such as ours) we want to pass the shaders down from the modelling step - this PR allows for that without changing the default behaviour from current.

## Additional info

@antirotor based on the discussion in the last PR and personal preference I switched the setting from `strip_shaders` to `include_shaders` and made the default False - let me know if you'd prefer it as a strip setting and I can update the PR.

## Testing notes:

Tested by exporting a model through Maya and ensuring that the model retains it's shaders with the option on - with existing behaviour when published without `include_shaders` on.

### DIFF
--- a/client/ayon_maya/plugins/create/create_model.py
+++ b/client/ayon_maya/plugins/create/create_model.py
@@ -39,5 +39,9 @@ class CreateModel(plugin.MayaCreator):
                     placeholder="attr1, attr2"),
             TextDef("attrPrefix",
                     label="Custom Attributes Prefix",
-                    placeholder="prefix1, prefix2")
+                    placeholder="prefix1, prefix2"),
+            BoolDef("include_shaders",
+                    label="Include Shaders",
+                    tooltip="Include shaders in the geometry export.",
+                    default=False),
         ]
--- a/client/ayon_maya/plugins/create/create_model.py
+++ b/client/ayon_maya/plugins/create/create_model.py
@@ -16,6 +16,7 @@ class CreateModel(plugin.MayaCreator):
 
     write_color_sets = False
     write_face_sets = True
+    include_shaders = False
 
     def get_instance_attr_defs(self):
 
@@ -43,5 +44,5 @@ class CreateModel(plugin.MayaCreator):
             BoolDef("include_shaders",
                     label="Include Shaders",
                     tooltip="Include shaders in the geometry export.",
-                    default=False),
+                    default=self.include_shaders),
         ]
--- a/client/ayon_maya/plugins/publish/extract_model.py
+++ b/client/ayon_maya/plugins/publish/extract_model.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """Extract model as Maya Scene."""
 import os
+from contextlib import nullcontext
 
 from ayon_core.pipeline import publish
 from ayon_maya.api import lib
@@ -69,6 +70,10 @@ class ExtractModel(plugin.MayaExtractorPlugin,
                           noIntermediate=True,
                           long=True)
 
+        # Check if shaders should be included as part of the model export. If
+        # False, the default shader is assigned to the geometry.
+        include_shaders = instance.data.get("include_shaders", False)
+
         with lib.no_display_layers(instance):
             with lib.displaySmoothness(members,
                                        divisionsU=0,
@@ -76,8 +81,11 @@ class ExtractModel(plugin.MayaExtractorPlugin,
                                        pointsWire=4,
                                        pointsShaded=1,
                                        polygonObject=1):
-                with lib.shader(members,
-                                shadingEngine="initialShadingGroup"):
+                with (
+                    nullcontext()
+                    if include_shaders
+                    else lib.shader(members, shadingEngine="initialShadingGroup")
+                ):
                     with lib.maintained_selection():
                         cmds.select(members, noExpand=True)
                         cmds.file(path,
--- a/client/ayon_maya/plugins/publish/extract_obj.py
+++ b/client/ayon_maya/plugins/publish/extract_obj.py
@@ -2,12 +2,14 @@
 import os
 
 import pyblish.api
+from ayon_core.pipeline import OptionalPyblishPluginMixin
 from ayon_maya.api import lib
 from ayon_maya.api import plugin
 from maya import cmds
 
 
-class ExtractObj(plugin.MayaExtractorPlugin):
+class ExtractObj(plugin.MayaExtractorPlugin,
+                 OptionalPyblishPluginMixin):
     """Extract OBJ from Maya.
 
     This extracts reproducible OBJ exports ignoring any of the settings
@@ -28,6 +30,8 @@ class ExtractObj(plugin.MayaExtractorPlugin):
     }
 
     def process(self, instance):
+        if not self.is_active(instance.data):
+            return
 
         # Define output path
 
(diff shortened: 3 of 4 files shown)


client/ayon_maya/plugins/create/create_model.py
```diff
@@ -39,5 +39,9 @@ def get_instance_attr_defs(self):
                     placeholder="attr1, attr2"),
             TextDef("attrPrefix",
                     label="Custom Attributes Prefix",
-                    placeholder="prefix1, prefix2")
+                    placeholder="prefix1, prefix2"),
+            BoolDef("include_shaders",
```
nikshankarnoble, line 43: @antirotor should this be `"include_shaders"` or `"includeShaders"`? I stuck with snake case to match your PR, but it seems that camelCase is the convention for the other settings.

BigRoy: In this case the snake case is good. Thanks for checking! :)

LiborBatek (approved):
Default state been correctly as ```Disabled``` which is preferable.

Screenshot 2024-08-19 164924

All working fine for both states of ```Include shaders``` settings...

When disabled, only mesh without shaders been produced.

When enabled, shaders are included in model with suitable representation like maya ascii or FBX.

Screenshot 2024-08-19 165822

and when the ```model``` product been loaded and shader present too:

Screenshot 2024-08-19 165748

client/ayon_maya/plugins/create/create_model.py
```diff
@@ -39,5 +39,9 @@ def get_instance_attr_defs(self):
...
+                    label="Include Shaders",
+                    tooltip="Include shaders in the geometry export.",
+                    default=False),
```
BigRoy, line 46: @LiborBatek looking at this code I have a feeling the setting in settings does not influence the default state here at all?

BigRoy: It should take it from a class attribute `self.include_shaders` similar to this: https://github.com/ynput/ayon-maya/pull/69/files#diff-d3f80c6626121c1d853dcc382eb66d18761a2aad4b6d7028125857fa5e552befR30

Of course `include_shaders = False` should also be added as class attribute.

```suggestion
                    default=self.include_shaders),
```

antirotor: I agree - @nikshankarnoble I tried to fix it but I don't have permissions to push to your fork. Could you just make those changes? With that, it is good to be merged. Thanks!

antirotor (dismissed):
apart that default value, it looks great

LiborBatek:
@BigRoy @antirotor yep, it was the case...I had to enable it manually (but as I was testing with already existing ```model``` instance I was thinking it was in its default disabled state due to the fact it was created before the option been available aka in this code branch) - I wasnt aware of this malfunction then...

Great catch!

client/ayon_maya/plugins/create/create_model.py
BigRoy: I was able to push. ✅

BigRoy:
> @BigRoy @antirotor yep, it was the case...I had to enable it manually (but as I was testing with already existing `model` instance I was thinking it was in its default disabled state due to the fact it was created before the option been available aka in this code branch) - I wasnt aware of this malfunction then...
> 
> Great catch!

@LiborBatek could you try again with latest commit?

LiborBatek (changes requested):
Works well (Include Shader ON) for ```.ma``` , ```FBX```

but not working with ```OBJ``` repre type....it lacks the shader assignments

image


note: when creating fresh new ```model``` publish instances the ```Include shader``` and its state correctly get propagated (enabled/disabled)...so all good on this front

client/ayon_maya/plugins/create/create_model.py
nikshankarnoble: Lovely - thanks very much both!

LiborBatek:
How about that ```OBJ``` and included shaders on?? Are we resolving it? As they arent included when publish happens...

Otherwise all working nicely!

antirotor:
> How about that `OBJ` and included shaders on?? Are we resolving it? As they arent included when publish happens...
> 
> Otherwise all working nicely!

If I remember correctly from the previous PR it didn't work for OBJ even when you are setting the options. Might be bug in OBJ exporter in Maya

BigRoy:
@LiborBatek this took me way longer to debug why it didn't work until I realized in `.obj` the materials are stored in a separate `.mtl` file on export. So the export does work - but the integrated files do not work for `.obj`.

I'd say - it's a bit out of scope for this PR to make it work - but the issue is essentially:

1. We need to also integrate the `.mtl` file for the obj.
2. We need to ALSO change contents in the `.obj` file because it'd store e.g. a line like: `mtllib modelMain.mtl` which would then need to be changed to the published integrated filename for the `.mtl` file (unless we transfer the file without renaming, almost like a 'resource' and make sure it lives next to the `.obj` then it'd work).

mkolar:
I will play devils advocate here and ask....shouldn't we finally deprecated .obj in the default completely? is that even the right format to use nowadays?

LiborBatek:
```OBJ``` stands for something like "if any other method fails" use it! Its quite common to bring OBJ with material groups into Substance Painter, ZBrush etc. So I think we should keep it (on a pretty basic level)...

@BigRoy I would go with ```mat``` file as a resource file without renaming when integrate it...making things much easier! and still working as expected

BigRoy:
> I would go with mat file as a resource file without renaming when integrate it...making things much easier! and still working as expected

Fine with me - but.... this PR or separate PR?

BigRoy:
> I would go with `mat` file as a resource file without renaming when integrate it...making things much easier! and still working as expected

Added to this PR.
@LiborBatek please test again.

LiborBatek (changes requested):
Now it does produce ```OBJ``` with accompanying ```mat``` file but it lacks texure connections (not sure if we can tackle it easily)

here is my ```OBJ``` model loaded (aka imported) into clean scene and also inspected resources:
Screenshot 2024-08-28 150841

published repre:

Screenshot 2024-08-28 150859

and when inspecting ```mat``` file in text editor:

Screenshot 2024-08-28 150915

BigRoy:
> file but it lacks texure connections (not sure if we can tackle it easily)

Does this work with regular `.obj` export for you? This may be a limitation of the `.obj` format for certain shaders. Can you confirm whether this is due to how we export it for AYON or just happens to be an issue with OBJ in general? @LiborBatek

LiborBatek:
Small note: I also lack the ```OBJ``` extractor knob when set in Ayon Settings as ```Optional``` it is not just present in the ```Publisher``` UI...

image

BigRoy:
> Small note: I also lack the `OBJ` extractor knob when set in Ayon Settings as `Optional` it is not just present in the `Publisher` UI...
> 
> image

Fixed with https://github.com/ynput/ayon-maya/pull/69/commits/f04a0c26715d146bd5dfd8e440af368060e9c77d

LiborBatek (changes requested):
Now it correctly shows ```Extract OBJ``` as optional within Publisher UI..so this been fixed!

Screenshot 2024-08-28 153933

Now also the ```mat``` file contains the textures inside...but we miss the actual image files copied next to the ```OBJ``` file...as when doing so manually (File>Export selection (OBJ))

Screenshot 2024-08-28 154119

here is the result when done manually:



Screenshot 2024-08-28 153230

BigRoy:
I really am starting to hate the `.obj` and `.mtl` file format.
Please try again. :)

LiborBatek (approved):
Now all works as expected and also ```OBJ``` representation having shader and corresponding source files / textures included.

All good! I think its good to go!

Screenshot 2024-08-29 095812

Also tested by importing the ```OBJ``` into empty workfile...

BigRoy:
@mkolar will hate me - but I'm merging this :) We can fight later.

PR #70, opened 2024-08-19
Validate unique names only within the instance not in full scene

BigRoy:
## Changelog Description

Validate unique names only within the instance not in full scene

## Additional info

This is allowed:
image

This is not allowed:
image

The validation only checks for member transforms that are included in the export instance.

image

## Testing notes:

Make sure to enable the validator: `ayon+settings://maya/publish/ValidateUniqueNames/enabled`

1. Create a scene with non-unique node names (e.g. see screenshots above).
2. When both included in the instance it should not be allowed, 
3. Otherwise when just present elsewhere in the scene unrelated to the export it should be allowed.

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_unique_names.py
+++ b/client/ayon_maya/plugins/publish/validate_unique_names.py
@@ -1,3 +1,5 @@
+from collections import Counter
+
 import ayon_maya.api.action
 from ayon_core.pipeline.publish import (
     OptionalPyblishPluginMixin,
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y:
I recommend testing with multiple instances present. 

I also wanted to make some backward compatibility settings so you could turn on the old behavior, but it seems that it is not actually necessary and new behaviour is quite logical.

BigRoy:
> I recommend testing with multiple instances present.
> 
> I also wanted to make some backward compatibility settings so you could turn on the old behavior, but it seems that it is not actually necessary and new behaviour is quite logical.

Confirmed:

image

image

Also works.

PR #74, opened 2024-08-22
Implementation of layout loader from Unreal to Maya

### LINKED ISSUE #65: Load layout from JSON

Currently, Maya can load only layout published from Maya (as .ma file). But other DCCs like Blender or Unreal are publishing Layout as JSON file.

Implement similar logic as is in other DCCs to pull in JSON. It would be also great if it add functionality described here https://github.com/ynput/ayon-unreal/issues/6

_No response_

### Additional context:

_No response_

moonyuet:
## Changelog Description
Support loading layout from Unreal to Maya
Resolve #65

## Additional info
Tested with the published layout from https://github.com/ynput/ayon-unreal/pull/109
Currently not supports updating non-existing loading assets.

## Testing notes:

1. Publish Layout from Unreal
2. Load Layout in Maya


LiborBatek (changes requested):
Did try to ```Load (JSON)``` and got following error msg:

```
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_0.4.5-dev.1\ayon_core\tools\loader\models\actions.py", line 740, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_0.4.5-dev.1\ayon_core\pipeline\load\utils.py", line 325, in load_with_repre_context
  File "c:\Work\REPO\ayon-maya\client\ayon_maya\plugins\load\load_layout.py", line 181, in load
    self._process(path, options)
  File "c:\Work\REPO\ayon-maya\client\ayon_maya\plugins\load\load_layout.py", line 145, in _process
    self.set_transformation(instance_name, transform)
  File "c:\Work\REPO\ayon-maya\client\ayon_maya\plugins\load\load_layout.py", line 148, in set_transformation
    asset = [
IndexError: list index out of range
```

image

but seems like one asset got in...

LiborBatek (changes requested):
All assets gets in my workfile but the transforms arent transferred correctly and all asssets been at zero coords

Screenshot 2024-09-18 133954

Im also enclosing the log for more info

[log.txt](https://github.com/user-attachments/files/17043531/log.txt)

LiborBatek (changes requested):
Now the products been all loaded but with wrong transforms.

Here I have ue map with loaded layout:

Screenshot 2024-09-19 175738

and in maya


Screenshot 2024-09-20 090451

and one particular asset transforms:


Screenshot 2024-09-20 090958

and in maya:


Screenshot 2024-09-20 090722

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+
+        all_loaders = discover_loader_plugins()
+
+        if not loaded_options:
+            loaded_options = []
+
+        repre_entities_by_version_id = self._get_repre_entities_by_version_id(
+            data
+        )
+        for element in data:
```
BigRoy, line 92: Can we get some hints or comments on what we expect `data` to be so that just by reading the code we can have some expectancy of what we assume is in the json?

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+                if not repre_entities:
+                    self.log.error(
+                        f"No valid representation found for version"
+                        f" {version_id}")
+                    continue
+                repre_entity = repre_entities[0]
+                repre_id = repre_entity["id"]
+                repr_format = repre_entity["name"]
+
+
```
BigRoy, line 107:
```suggestion
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+            if version_id:
+                repre_entities = repre_entities_by_version_id[version_id]
+                if not repre_entities:
+                    self.log.error(
+                        f"No valid representation found for version"
+                        f" {version_id}")
+                    continue
+                repre_entity = repre_entities[0]
+                repre_id = repre_entity["id"]
+                repr_format = repre_entity["name"]
```
BigRoy, line 105: Any reason we just assume it's always the first representation?

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+                    instance_name = instance_name.replace(f"_{extension}", "")
+                    self.set_transformation(instance_name, transform)
+
+    def set_transformation(self, instance_name, transform):
+        asset = self.get_asset(instance_name)
+        translation = [
+            transform["translation"]["x"],
+            transform["translation"]["z"],
+            transform["translation"]["y"]
+            ]
```
BigRoy, line 167:
```suggestion
        ]
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+
+        self.log.info(">>> loading json [ {} ]".format(path))
+        self._process(path, options)
+        folder_name = context["folder"]["name"]
+        namespace = namespace or unique_namespace(
+            folder_name + "_",
+            prefix="_" if folder_name[0].isdigit() else "",
+            suffix="_",
+        )
+        # TODO: implement the function to load all the assets and set transforms
```
BigRoy, line 199: Is this todo still relevant?

```suggestion
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+    def update(self, container, context):
+        repre_entity = context["representation"]
+        path = get_representation_path(repre_entity)
+        self._process(path, options=None)
+        # Update metadata
+        node = container["objectName"]
+        cmds.setAttr("{}.representation".format(node),
+                     repre_entity["id"],
+                     type="string")
+        self.log.info("... updated")
```
BigRoy, line 216:
```suggestion
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def switch(self, container, context):
+        self.update(container, context)
+
+    def remove(self, container):
+        members = cmds.sets(container['objectName'], query=True)
```
BigRoy, line 222: Note: if the container is empty this may return `None`.

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def _process(self, filepath, options, loaded_options=None):
+
+        with open(filepath, "r") as fp:
+            data = json.load(fp)
+
+        all_loaders = discover_loader_plugins()
+
+        if not loaded_options:
+            loaded_options = []
```
BigRoy, line 87:
```suggestion
        if loaded_options is None:
            loaded_options = []
```

client/ayon_maya/plugins/load/load_layout.py
BigRoy, line 92: I wonder if this turns way more readable if we turn this element into a dedicated data structure here so that we know exactly which data to expect.

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+                        options=options
+                    )
+                instances = [
+                    item for item in data
+                    if ((item.get('version') and
+                        item.get('version') == element.get('version')))]
+
+                for instance in instances:
+                    transform = instance.get('transform')
+                    instance_name = instance.get('asset_name')
```
BigRoy, line 156:
```suggestion
                    transform = instance["transform"]
                    instance_name = instance["asset_name"]
```
Logic below would fail anyway if these were None, so it seems like we require these to exist.

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,231 @@
...
+            name=name,
+            namespace=namespace,
+            nodes=[],
+            context=context,
+            loader=self.__class__.__name__)
+
+    def update(self, container, context):
+        repre_entity = context["representation"]
+        path = get_representation_path(repre_entity)
+        self._process(path, options=None)
```
BigRoy, line 210: Pretty sure updates won't work here - this will just load everything AGAIN on update?

moonyuet: it won't load everything again. it would only update on the transform. There is a check on whether the loaded container exists before loading the assets.

BigRoy: But what if objects were removed from layout? Or if new ones were added?

Also - I think the code may become more readable if somehow you can have a `process_element` or whatever 'element' is so you can reduce the nesting a bit. Like along the lines of:
```python
for element in data:
    self.process_element(element)
```

moonyuet: > But what if objects were removed from layout? Or if new ones were added?

this is something we should consider for not only the loader here but also the one lasted in unreal

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,233 @@
...
+from ayon_maya.api.pipeline import containerise
+
+
+class LayoutLoader(plugin.Loader):
+    """Layout Loader(json)"""
+
+    product_types = {"layout"}
+    representations = {"json"}
+
+    label = "Layout Loader(json)"
```
BigRoy, line 25: Pretty sure the `(json)` is redundant because it will show added to the label in the loader anyway?
```suggestion
    label = "Load Layout"
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,233 @@
...
+            if loader.__name__ == name:
+                return loader
+
+        return None
+
+    def get_asset(self, instance_name):
+        container = [
+            con for con in cmds.ls(f"{instance_name}*")
+            if con.endswith("_CON")
+        ][0]
```
BigRoy, line 75: Isn't this the same as this?
```suggestion
        container = next(cmds.ls(f"{instance_name}*_CON"))
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+        for element in data:
+            repre_id = None
+            repr_format = None
+            version_id = element.get("version")
+            if version_id:
+                repre_entities = repre_entities_by_version_id[version_id]
+                if not repre_entities:
+                    self.log.error(
+                        f"No valid representation found for version"
```
BigRoy, line 102:
```suggestion
                        "No valid representation found for version"
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,233 @@
...
+            # If reference is None, this element is skipped, as it cannot be
+            # imported in Maya
+            if not repre_id:
+                continue
+
+            instance_name = element.get('instance_name')
+            containers = [
+                con for con in cmds.ls(f"{instance_name}*")
+                if con.endswith("_CON")
+            ]
```
BigRoy, line 121: I'm honestly quite confused by this logic - what's this "lookup" by name? Why?

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,233 @@
...
+                                    if repre_entity["name"] == extension), None)
+                repre_id = repre_entity["id"]
+                repr_format = repre_entity["name"]
+
+            # If reference is None, this element is skipped, as it cannot be
+            # imported in Maya
+            if not repre_id:
+                continue
+
+            instance_name = element.get('instance_name')
```
BigRoy, line 117:
```suggestion
            instance_name: str = element["instance_name"]
```

client/ayon_maya/plugins/load/load_layout.py
BigRoy: Also, this feels **very** insecure from a pipeline standpoint to just assume the first entry by pure luck is the one we need here?

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,233 @@
...
+    def get_asset(self, instance_name):
+        container = [
+            con for con in cmds.ls(f"{instance_name}*")
+            if con.endswith("_CON")
+        ][0]
+        namespace = cmds.getAttr(f"{container}.namespace")
+        asset = [asset for asset in cmds.ls(f"{namespace}:*", assemblies=True)][0]
+        return asset
+
+    def _process_element(self, filepath, options, loaded_options=None):
```
BigRoy, line 80: Function name seems wrong. This loads json data and then continues to iterate over `elements` in `data`. So it does not process a single element, it processes the full json?

This would be processing the element:
```python
for element in data:
    self._process_element(element)
...

def _process_element(self, element):
    # process the element
```

BigRoy (commented):
Some comments, but also lacks docstrings on the methods - it's quite hard to grasp what certain things are trying to do? E.g. why do we rely on the first match with `{instance_name}*_CON` to be "the asset" (whatever that is?)

client/ayon_maya/plugins/load/load_layout.py
moonyuet: It's looking up the container of the loaded assets. If the loaded asset is there, it won't be loaded again.

moonyuet:
> Some comments, but also lacks docstrings on the methods - it's quite hard to grasp what certain things are trying to do? E.g. why do we rely on the first match with `{instance_name}*_CON` to be "the asset" (whatever that is?)

we need to load the asset and set the transformation. Similar thing happens in the unreal layout loader. 
I will improve the code as the current one can't support the same asset loaded multiple of times.

client/ayon_maya/plugins/load/load_layout.py
moonyuet: it would be great if you can have better solutions to do this since the unreal layout loader did it similarly to load the asset and set the transformation.

client/ayon_maya/plugins/load/load_layout.py
moonyuet: Maybe it is wrong but there is a parsed data in the representation by version id. This is something we should also be concerned about.

BigRoy:
> Maybe it is wrong but there is a parsed data in the representation by version id. we need to load the asset and set the transformation. Similar thing happens in the unreal layout loader.
> I will improve the code as the current one can't support the same asset loaded multiple of times.

_What_ is "the asset"? Do you just happen to mean the root node of whatever is being loaded?

client/ayon_maya/plugins/load/load_layout.py
BigRoy: > Maybe it is wrong but there is a parsed data in the representation by version id. This is something we should also be concerned about.

Could you point to where that's at? I don't think I understand what "parsed data" is problematic.

client/ayon_maya/plugins/load/load_layout.py
BigRoy: But - how would that 'container' magically match exactly the name from the JSON data?

client/ayon_maya/plugins/load/load_layout.py
moonyuet: > > Maybe it is wrong but there is a parsed data in the representation by version id. This is something we should also be concerned about.
> 
> Could you point to where that's at? I don't think I understand what "parsed data" is problematic.

I will improve the code later, I believe the code above provides me some examples for what I should improve.

moonyuet:
> > Maybe it is wrong but there is a parsed data in the representation by version id. we need to load the asset and set the transformation. Similar thing happens in the unreal layout loader.
> > I will improve the code as the current one can't support the same asset loaded multiple of times.
> 
> _What_ is "the asset"? Do you just happen to mean the root node of whatever is being loaded?



> > Maybe it is wrong but there is a parsed data in the representation by version id. we need to load the asset and set the transformation. Similar thing happens in the unreal layout loader.
> > I will improve the code as the current one can't support the same asset loaded multiple of times.
> 
> _What_ is "the asset"? Do you just happen to mean the root node of whatever is being loaded?

If the root nodes means the certain objects from other product type which are marked in the json, yes.

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,223 @@
...
+                return loader
+
+        return None
+
+    def get_asset(self, containers, instance_name):
+        asset = None
+        containers = [con for con in containers if con.startswith(instance_name)]
+        for container in containers:
+            namespace = cmds.getAttr(f"{container}.namespace")
+            asset = [asset for asset in cmds.ls(f"{namespace}:*", assemblies=True)][0]
```
BigRoy, line 76: If there's one thing I've learned from Maya is that even the namespaces are hard to rely on :') hehe.

Anyway, this code is the same as:
```suggestion
        for container in containers:
            namespace = cmds.getAttr(f"{container}.namespace")
            asset = next(iter(cmds.ls(f"{namespace}:*", assemblies=True)))
```

However that would fail as soon as someone had "reparented" the object in their scene locally and it is not 'top level assembly anymore'.

We might want to get the highest root from the container. Also, why do we need to iterate the containers (when would the first container fail?) If there're edge cases there it may be good to document them as comments in the code to explain the reasoning.

Getting the highest root of the container may be someting like this:
```suggestion
        # Get the highest root node from the loaded container
        for container in containers:
            members = lib.get_container_members(container)
            roots = lib.get_highest_in_hierarchy(members)
            root = next(iter(roots), None)
            if root is not None:
                return root
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,223 @@
...
+        if version_id:
+            repre_entities = repre_entities_by_version_id[version_id]
+            if not repre_entities:
+                self.log.error(
+                    "No valid representation found for version"
+                    f" {version_id}")
+                return
+            # always use the first representation to load
+            # If reference is None, this element is skipped, as it cannot be
+            # imported in Maya, repre_entities must always be the first one
```
BigRoy, line 92: I understand you added this comment based on my PR feedback - however, the comment does not explain **why** we do this? WHY is always using the first representation the 'good' choice? (What if for whatever reason the first representation is an image or thumbnail or whatever?)

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,223 @@
...
+        Args:
+            transform (list): Transformations of the asset
+            rotation (list): Rotations of the asset
+
+        Returns:
+            List[om.MMatrix]: List of transformation matrix of the asset
+        """
+        transform_mm = om.MMatrix(transform)
+        convert_transform = om.MTransformationMatrix(transform_mm)
+        print(rotation)
```
BigRoy, line 154:
```suggestion
```

client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -0,0 +1,223 @@
...
+from ayon_maya.api.pipeline import containerise
+
+
+class LayoutLoader(plugin.Loader):
+    """Layout Loader (json)"""
+
+    product_types = {"layout"}
+    representations = {"json"}
+
+    label = "Layout Loader"
```
BigRoy, line 25: As far as I know most "load" entries in are either `Load X`, `Import X` or `Reference X`.
```suggestion
    label = "Load Layout"
```

Generating some consistency in those actions has come up before - e.g. [here](https://github.com/ynput/OpenPype/issues/694#issuecomment-1091081271) but I'm not sure if @mkolar @antirotor has new notes about that.

client/ayon_maya/plugins/load/load_layout.py
moonyuet: it is only applied for the layout published from maya and re-loaded back to maya, it errors out for the repre_id

client/ayon_maya/plugins/load/load_layout.py
moonyuet: That's my bad. I will just rename it back to the Load Layout

client/ayon_maya/plugins/load/load_layout.py
BigRoy: Which raises another question - why does that fail, etc? Like _what_ is the use case? And do we think that's completely fine? Are we lacking some data? This seems to be hiding a bug in plain sight? What am I missing?

moonyuet: > Which raises another question - why does that fail, etc? Like _what_ is the use case? And do we think that's completely fine? Are we lacking some data? This seems to be hiding a bug in plain sight? What am I missing?

But it works when I loaded the layout from unreal, it works. Maybe I need to test it multiple times to ensure if there is anything missing.

LiborBatek (changes requested):
Still doesnt work for me...assets got in but no transforms applied...

Screenshot 2024-10-21 164426

And in maya:
Screenshot 2024-10-21 164326

And yes, I have checked out those branches for both ayon-unreal and ayon-maya...

LiborBatek (changes requested):
It works sort of...speaking of assets and transforms, we just miss negative Scale in X axis.

We should apply it on an asset level and in world coord system I guess, see my comparison images:

Original UE layout:

Screenshot 2024-10-23 134705

and from TOP view:

Screenshot 2024-10-23 134619

Once loaded into maya (all transforms besides rotation seems correct):


Screenshot 2024-10-23 134743


Screenshot 2024-10-23 134805

So to fix rotations I have added locator for applying Scale -X in world coords resulting into matching layout to UE:


Screenshot 2024-10-23 140839

and here a schema for that adjustment:

Screenshot 2024-10-23 141006

LiborBatek:
@moonyuet nevermind my Scale value of ```100``` it was just so the locator having bigger size in the viewports

LiborBatek:
I think just adding 90 deg rot Y cant solve that mirror issue tbh

not working...
image

moonyuet:
> I think just adding 90 deg rot Y cant solve that mirror issue tbh
> 
> not working... image

Can you send me the robot model? I believe we can get the correct transformation data through xform rather than adding negative scale to the data

LiborBatek:
this is what I have got atm when tested with latest commits:

UE:


Screenshot 2024-10-29 165529


Maya:

Screenshot 2024-10-29 164807

LiborBatek:
its a bit different this time but still not correct...

image

LiborBatek (approved):
Hooray! It finally works and produces 1:1 correct layout in maya from UE!!

I have rather performed another test with different layout and all working well and nicely!!

Congrats on this one!
Screenshot 2024-10-30 133714

Looks perfect to me. And also checked properly so it really does :)

antirotor:
Just a little question -  did we try it  with blender layouts @LiborBatek ? They should have json too? It should work I suppose, but just asking.

antirotor (approved):

LiborBatek:
> Just a little question - did we try it with blender layouts @LiborBatek ? They should have json too? It should work I suppose, but just asking.

nope hasnt been tested in blender, and I also think ```blender``` was not taken into account for that? @moonyuet or?

moonyuet:
> > Just a little question - did we try it with blender layouts @LiborBatek ? They should have json too? It should work I suppose, but just asking.
> 
> nope hasnt been tested in blender, and I also think `blender` was not taken into account for that? @moonyuet or?

Let's test it in blender maybe? The layout product is originally implemented between blender and unreal. So would be good to know if any error occurred in Blender. But if there is an issue for blender, it should be fixed in blender host.

antirotor:
definitely, it was just a question, not suggestion to fix it here :) If this works (and it looks so) I am happy to merge and deal with maya <-> blender layout in different time and place.

PR #75, opened 2024-08-22
Improve applying render resolution and aspect ratio on render settings reset

BigRoy:
## Changelog Description

Fix pixel aspect ratio / device aspect ratio getting messed up for Arnold renderer on render settings reset.

Additionally:
- This now applies the resolution from the task entity, not the folder entity.
- This now also applies pixel aspect ratio as defined on the entity.

## Additional info

This fixes a bug when resetting render resolution for Arnold where on reset the pixel/device aspect ratio would be oddly set with a pixel aspect ratio that is not 1.0.

E.g. left is before, right is after resetting render settings **without this PR**
image

After this PR the pixel aspect ratio will now adhere to what is specified on the task entity.
_This bug was only occurring if the current renderer render settings were reset for was actively Arnold (mtoa) renderer._

## Testing notes:

1. Resetting render settings should work for the different renderers and should set the resolution (and pixel aspect ratio correctly!)
    - [x] Arnold (mtoa)
    - [x] Redshift (should work if Arnold works because it uses same attributes as Arnold)
    - [x] V-Ray (may be good to check because it uses a custom render settings node!)
    - [ ] Renderman (should work if Arnold works because it uses same attributes as Arnold)

Note that this is about **Set Render Settings** or the feature that it automatically does so on creating a render publish instance in the scene (enabled by default).

image

### DIFF
--- a/client/ayon_maya/api/lib_rendersettings.py
+++ b/client/ayon_maya/api/lib_rendersettings.py
@@ -7,7 +7,7 @@ from ayon_core.lib import Logger
 from ayon_core.settings import get_project_settings
 
 from ayon_core.pipeline import CreatorError, get_current_project_name
-from ayon_core.pipeline.context_tools import get_current_folder_entity
+from ayon_core.pipeline.context_tools import get_current_task_entity
 from ayon_maya.api.lib import reset_frame_range
 
 
@@ -77,8 +77,6 @@ class RenderSettings(object):
             renderer = cmds.getAttr(
                 'defaultRenderGlobals.currentRenderer').lower()
 
-        folder_entity = get_current_folder_entity()
-        folder_attributes = folder_entity["attrib"]
         # project_settings/maya/create/CreateRender/aov_separator
         try:
             aov_separator = self._aov_chars[(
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
Works for ```Arnold``` ```Rshift``` and probably for ```Renderman``` 

but fails with ```Vray``` throwing following error msg:

```
# Error: TypeError: file C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.2.11-dev.1\ayon_maya\api\lib_rendersettings.py line 113: _set_vray_settings() missing 3 required positional arguments: 'height', 'pixel_aspect', and 'aov_separator'
```

BigRoy:
> but fails with `Vray` throwing following error msg:
> 
> ```
> # Error: TypeError: file C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.2.11-dev.1\ayon_maya\api\lib_rendersettings.py line 113: _set_vray_settings() missing 3 required positional arguments: 'height', 'pixel_aspect', and 'aov_separator'
> ```

Sorry about that. Completely my bad! Please try again.

LiborBatek (approved):
All working now even with ```Vray```.

Also tested with overrides on ```Entity``` level and ```Task``` level and both are working nicely!

I guess we can merge it (saving some time with Renderman...as it should work same as Arnold)

PR #68, opened 2024-08-12
Add Extract Animation to ayon settings 

moonyuet:
## Changelog Description
Resolve #66 and #36
This PR is to add Extract Animation to the ayon setting, and use apply_setting function into ExtractAnimation to set enabled state.
image

## Additional info
Build maya addon with this branch

## Testing notes:

1. Build Maya addon with this branch
2. Install Maya addon
3. Launch Maya
4. Publish Animation Product Type.

PR #77, opened 2024-08-27
Fix settings for Maya USD Animation Extractor

BigRoy:
## Changelog Description

Fix name in settings to match with name of plug-in to ensure settings are actually applied

## Additional info

Originally reported by @LiborBatek here: https://github.com/ynput/ayon-maya/pull/68#issuecomment-2288791598

## Testing notes:

1. Settings should adjust the extractor's defaults
2. Disabling it in settings should hide the attributes from the publisher UI

### DIFF
--- a/server/settings/publishers.py
+++ b/server/settings/publishers.py
@@ -550,7 +550,7 @@ class ExtractMayaUsdPointcacheModel(BaseSettingsModel):
     active: bool = SettingsField(title="Active")
 
 
-class ExtractMayaUsdAnimationModel(BaseSettingsModel):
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
Code looks good for me. Haven't tested.

LiborBatek (approved):
Works well.

Been tested on ```model``` ```animation``` and ```pointcache``` publish instances...as soon as disabled extraction of ```USD``` all the options disappear from the publisher UI.
Screenshot 2024-08-28 165650



Screenshot 2024-08-28 165705

Issue #81, opened 2024-08-30
Collect Camera (FBX) is missing "Optional" and "Active" Settings in the addon settings


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

There is no these options as for ```Collect Animation (FBX)``` and ```Collect Model (FBX)``` which do have it.

<img width="404" alt="Screenshot 2024-08-30 113745" src="https://github.com/user-attachments/assets/84e26c68-353f-4252-9c58-13a097d5872a">

It would be good to have the ability to expose such settings and propagate into Publisher UI.

### How would you imagine the implementation of the feature?

It should behave as for other ```FBX``` collectors... we already have.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

PR #83, opened 2024-08-30
Add optional and active state for fbx camera collector in ayon setting

moonyuet:
## Changelog Description
Resolve https://github.com/ynput/ayon-maya/issues/81

## Additional info
Build maya addon with this branch and install it into ayon server

## Testing notes:

1. Build maya addon with this branch and install it into ayon server
2. You can choose whether you want to enable the fbx camera collector through `ayon+settings://maya/publish/CollectFbxCamera`
3. Launch Maya
4. Create Camera Instance
5. Publish

### DIFF
--- a/server/settings/publishers.py
+++ b/server/settings/publishers.py
@@ -180,10 +180,6 @@ class CollectFbxAnimationModel(BaseSettingsModel):
     enabled: bool = SettingsField(title="Collect Fbx Animation")
 
 
-class CollectFbxCameraModel(BaseSettingsModel):
-    enabled: bool = SettingsField(title="CollectFbxCamera")
-
-
 class CollectGLTFModel(BaseSettingsModel):
     enabled: bool = SettingsField(title="CollectGLTF")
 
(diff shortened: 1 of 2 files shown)


server/settings/publishers.py
```diff
@@ -1091,7 +1087,9 @@ class PublishersModel(BaseSettingsModel):
         "enabled": False
     },
     "CollectFbxCamera": {
-        "enabled": False
+        "enabled": False,
+        "optional": True,
```
BigRoy, line 1091: Previously it was not optional - should we preserve that behavior so that it's by default still not optional unless studio enables it? :)

LiborBatek (approved):
All working fine...when set in the Ayon Settings for maya addon, its visible in the Publisher UI as optional settings:

Screenshot 2024-08-30 141241

Has been also tested if enabling/disabling works for publishing and also loading those FBXs back to maya for check.

All working in any circumstances...

server/settings/publishers.py
LiborBatek: yeah, I do agree with you, that we should keep it as ```optional``` ```false``` as defaults (so the TDs can expose it if desired)

BigRoy (approved):

PR #85, opened 2024-08-31
Look Assigner: Ignore invalid entity ids in `get_folders` query 

### LINKED ISSUE #84: Look Assigner fails to list assets if entries have invalid folder id

### Current Behavior:

The Maya Look Assigner tool silently fails to load any asset entries as soon as it, in selection or in scene (depending on UI button you press) it goes over nodes that have an invalid folder id in the `cbId` attribute. Like for example a `cbId` still origination from OpenPype.

### Expected Behavior:

It should log a warning on invalid ids, but list the others.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Load some assets
2. Go to the `cbId` attribute. Remove some characters from before the `:` in the value.
3. Try the look assigner on that node - it doesn't list anything. (should log a warning instead!)
4. Now try that node together with nodes that do have valid ids - it still lists nothing. (should list instead, plus log warning)

### Relevant log output:

_No response_

### Additional context:

_No response_

BigRoy:
## Changelog Description

Fixes an issue where look assigner would not list anything if it had to list entries that had invalid ids (those that are NOT valid `uuid` patterns) and hence the `ayon_api.get_folders` query would fail.

## Additional info

Ignore invalid entity ids in `get_folders` query to avoid `GraphQlQueryFailed` exception which then also fails to return for the valid ids in the set.

Fixes #84

Note that due to other logic in the code a warning will still be logged for these invalid ids, which is as intended. For example:
```
// Warning: ayon_maya.tools.mayalookassigner.commands : Id found on 2 nodes for which no folder is found database, skipping 'ac991b60b4c5a70'
// Warning: ayon_maya.tools.mayalookassigner.commands : Id found on 2 nodes for which no folder is found database, skipping 'ac991b60b1eeb7fc773031a70'
```

## Testing notes:

1. Load some assets
2. Go to the `cbId` attribute. Remove some characters from before the `:` in the value.
3. Try the look assigner on that node - it doesn't list anything. (should log a warning instead!)
4. Now try that node together with nodes that do have valid ids - it still lists nothing. (should list instead, plus log warning)

OR

Load a scene with OpenPype assets together with a scene with AYON asset. Look assigner should still list entries for AYON.

### DIFF
--- a/client/ayon_maya/tools/mayalookassigner/commands.py
+++ b/client/ayon_maya/tools/mayalookassigner/commands.py
@@ -130,6 +130,12 @@ def create_items_from_nodes(nodes):
     project_name = get_current_project_name()
     folder_ids = set(id_hashes.keys())
 
+    # Ignore invalid ids
+    folder_ids = {
+        folder_id for folder_id in folder_ids
+        if ayon_api.utils.convert_entity_id(folder_id)
+    }
+
     folder_entities = ayon_api.get_folders(
         project_name, folder_ids, fields={"id", "path"}
     )


BigRoy:
@m-u-r-p-h-y can you run this through your demo scene with the alembic character? :)

LiborBatek (approved):
I was able to get ```look``` assigned even when my selection of products included some parts without correct ```cbId```. LGTM

image

note: I didnt test the optional test step with legacy ```OpenPype```  assets mixed up with ```AYON``` ones for time reasons as its very edge case and I suppose it will behave similar as when cbId missing (due to time consuming data preparation)

PR #94, opened 2024-09-05
Create Multi-shot layout: Add label only if not same as shot name.

BigRoy:
## Changelog Description

In recent versions of AYON the "label" is always set by default and matches the "name" whereas previously, "label" defaulted to an empty value. Hence, due to that update, this logic now needed to change to correctly detect whether a customized label was set.

## Additional info

Fixes an issue where the shot name in camera sequencer would be `shot (shot)` where the labels between the brackets matched the shot name, making it look odd.

## Testing notes:

1. Run create multi-shot layout
2. Have shots with and without customized labels
3. Only those with customized labels should have that between brackest like `{name} ({label})`


LiborBatek (approved):
LGTM!

before this PR:
Screenshot 2024-09-20 150032

and with this PR:


Screenshot 2024-09-20 150638

PR #98, opened 2024-09-10
Set instance frame range attribute defaults to current task entity attributes

BigRoy:
## Changelog Description

Allow frame range defaults for instances to be based on current task entity attributes

## Additional info

Enable the setting: `ayon+settings://maya/create/use_entity_attributes_as_defaults`
image

## Testing notes:

1. Enable/disable the setting `ayon+settings://maya/create/use_entity_attributes_as_defaults`
2. Set your maya timeline different than current task entity.
3. When the setting is enabled, newly created instances should default to frame range of current context task entity.
4. When the setting is disabled (default for backwards compatibility), the frame range will by default be based off of the timeline.

_This will be the case for all Creators using the `collect_animation_defs` function. Which is as far as I know all the product types with frame ranges exposed in the publisher UI currently, e.g.  `pointcache`, `animation`, `yeticache`, etc._

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -570,24 +570,47 @@ def pairwise(iterable):
     return zip(a, a)
 
 
-def collect_animation_defs(fps=False):
+def collect_animation_defs(fps=False, create_context=None):
     """Get the basic animation attribute definitions for the publisher.
 
+    Arguments:
+        fps (bool): Whether to include `fps` attribute definition.
+        create_context (CreateContext | None): When provided the context of
+            publisher will be used to define the defaults for the attributes.
+            Depending on project settings this may then use the entity's frame
+            range and FPS instead of the current scene values.
+
     Returns:
-        OrderedDict
+        List[NumberDef]: List of number attribute definitions.
 
     """
 
-    # get scene values as defaults
-    frame_start = cmds.playbackOptions(query=True, minTime=True)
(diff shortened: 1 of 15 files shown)


client/ayon_maya/api/lib.py
```diff
@@ -570,24 +570,47 @@ def pairwise(iterable):
...
-        query=True, animationStartTime=True
-    )
-    frame_end_handle = cmds.playbackOptions(query=True, animationEndTime=True)
-
-    handle_start = frame_start - frame_start_handle
-    handle_end = frame_end_handle - frame_end
+    use_entity_frame_range = False
+    if create_context is not None:
+        project_settings = create_context.get_current_project_settings()
+        print(project_settings["maya"]["create"])
```
iLLiCiTiT, line 591:
```suggestion
```

LiborBatek:
How about to change the settings name to something simpler (I know its just cosmetics but why not to stick to simpler name for clarity)

something like:

```Use task context for setting frame range``` or similar?

BigRoy:
> How about to change the settings name to something simpler (I know its just cosmetics but why not to stick to simpler name for clarity)
> 
> something like:
> 
> `Use task context for setting frame range`
> 
> or similar?

Haha - on my day off; not even the time to change it with merges this quick!
Feel free to create a backlog issue if you hate the wording on the setting label and we can follow up.

PR #103, opened 2024-09-11
Remove deprecated color management settings 

BigRoy:
## Changelog Description

Removes deprecated color management settings for Maya.

This is a tricky one, because this will be backwards compatible for everyone who had not previously EXPLICITLY enabled `ayon+settings://maya/imageio/workfile/enabled` because it would've been disabled before and in that case they were still using the deprecated settings. Now, the newer values will apply when enabled by default.

If the setting was overridden explicitly to disabled it will now log a message that the AYON Maya workfile color management is disabled.

## Additional info

This was "deprecated" since OpenPype 3.15.5.

Also removed the legacy OCIO config settings that were for Maya 2020 and older since those are Py2 anyway and are already unsupported.

## Testing notes:

1. Test whether the color management settings work and apply as you'd expect
2. Preferably you also test how this influences older projects from before and highlight where the issues may appear.


client/ayon_maya/api/lib.py
```diff
@@ -3417,36 +3379,27 @@ def _colormanage(**kwargs):
         except RuntimeError as exc:
             log.error(exc)
 
+    log.info(f"Setting Maya colorspace..")
```
BigRoy, line 3382:
```suggestion
    log.info("Setting Maya colorspace..")
```

BigRoy:
We're now using this in production on our end. 👍

moonyuet (approved):

Issue #100, opened 2024-09-11
Extract Layout: Representation Type


moonyuet:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

In the latest develop of maya, we can only choose to publish layout with fbx and abc. This is not good for the case where unreal layout loader uses level to load layout when the loaded asset are fbx and abc at the same time.

### How would you imagine the implementation of the feature?

We can just use one extractor for this case, and publish the transformation data by checking on the repre type of the loaded asset.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

PR #104, opened 2024-09-11
Extract Layout: Add repre_type as json element 

moonyuet:
## Changelog Description
This PR is to add extension of the loaded asset as part of the elements in the json published from extract layout.
Resolve https://github.com/ynput/ayon-maya/issues/100

## Additional info
The adjustment of the codes does not mean the transformation data of alembic has been fixed (The related code still hasn't been touched)

## Testing notes:

1. Extract Layout
2. Should be published successfully


LiborBatek (approved):
Did produce new ```Layout``` and compared it to the older JSON and seems working nicely...

Screenshot 2024-09-12 093544

also loads normally into UE

Screenshot 2024-09-12 093817

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -99,13 +96,15 @@ def process(self, instance):
                 "instance_name": cmds.getAttr(
                     "{}.namespace".format(container)),
                 "representation": str(representation_id),
-                "version": str(version_id)
+                "version": str(version_id),
+                "extension": representation["name"]
```
BigRoy, line 100: Note that the representation name does not have to be the extension. A representation may also have `representation["ext"]`, right @iLLiCiTiT ? Not sure if that is ALWAYS set though.

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -181,15 +193,8 @@ def convert_transformation_matrix(self, transform_mm: om.MMatrix, rotation: list
         return convert_transform.asMatrix()
```
BigRoy, line 195:
```suggestion
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -150,14 +149,25 @@ def process(self, instance):
...
+                matrix = self.convert_abc_transformation_matrix(matrix, local_rotation)
+        else:
+            raise RuntimeError(
+                f"No convert transformation matrix found for the loaded asset in {extension}")
         t_matrix = convert_matrix_to_4x4_list(matrix)
         return t_matrix
 
-    def convert_transformation_matrix(self, transform_mm: om.MMatrix, rotation: list) -> om.MMatrix:
-        """Convert matrix to list of transformation matrix for Unreal Engine import.
+    def convert_abc_transformation_matrix(self, transform_mm: om.MMatrix, rotation: list) -> om.MMatrix:
```
BigRoy, line 169: This method and `convert_transformation_matrix` appear to be the exact same implementation? They don't differ, or am I blind? :)

And if they _were_ to differ - why do they need to?

moonyuet: I know you would ask this - There are some extreme cases of the orientation in terms of alembic import not resolved yet so that's why I left the function like this. If we can any idea to resolve this, and we can go back just edit the function.

client/ayon_maya/plugins/publish/extract_layout.py
moonyuet: I guess it's not supported yet?
image

iLLiCiTiT: @moonyuet the error is not because it is not supported, but because you don't fill it. You can fill `"ext"` with extension and `"name"` with whatever you want.

NOTE: I don't know full context, let me know If I'm off.

iLLiCiTiT: I just looked and you're working with existing representation entity. So the ext is in `representation["context"]["ext"]`.

NOTE: And also found out that it takes product type from representation context too, which is not really good idea. It should fetch product entity and use `productType` from it.

PR #107, opened 2024-09-13
Look: Collect component assignment nodes as shapes

### LINKED ISSUE #106: Look: Validate Look Sets fail if mesh has only face assignments

### Current Behavior:

When publishing a look on a mesh that for whatever reason has the looks assigned to all the components instead of the shape itself the validation will log it as 'invalid'.

image

Why does this happen?

This is because querying the members for a shader when there are face assignments Maya returns those like `pCube1.f[0:200]` instead of `pCubeShape1.f[0:200]` so it returns the transform - instead of the shape. Which is wildly confusing.

Anyway, we may want to, on collect, if the member has components and the node is a transform collect the shape instead. That may actually even fix potential issues where we're accidentally collecting the ids from the transform instead of the shape?

### Expected Behavior:

It should either be allowed (if it doesn't provide issues elsewhere) or provide a clearer validation error so the user knows how to fix it.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Select all faces of a mesh.
2. Assign a material
3. Publish

(This is just to make an easy reproducable; however, similar issues may arise on merging or separating meshes or other maya construction history)

### Relevant log output:

```shell
ERROR: Missing '|cube|cubeShape' in collected set node 'blinn1SG'
  File "C:\Users\Maqina-RTX-03\AppData\Local\Ynput\AYON\dependency_packages\ayon_2407291312_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\Maqina-RTX-03\AppData\Local\Ynput\AYON\addons\maya_0.2.12-cb.1\ayon_maya\plugins\publish\validate_look_sets.py", line 47, in process
    raise PublishValidationError(
ayon_core.pipeline.publish.publish_plugins.PublishValidationError: 'lookMain' has relationships that could not be collected, likely due to lack of a `cbId` on the relevant nodes or sets.
```

### Additional context:

_No response_

BigRoy:
## Changelog Description

Fix nodes with component assignments not being collected as the shape but the transform instead

## Additional info

Fixes #106 

## Testing notes:

1. Select all faces of a mesh.
2. Assign a material
3. Publish

(This is just to make an easy reproducable; however, similar issues may arise on merging or separating meshes or other maya construction history)

### DIFF
--- a/client/ayon_maya/plugins/publish/collect_look.py
+++ b/client/ayon_maya/plugins/publish/collect_look.py
@@ -486,8 +486,12 @@ class CollectLook(plugin.MayaInstancePlugin):
             dict
 
         """
-
         node, components = (member.rsplit(".", 1) + [None])[:2]
+        if components and not cmds.objectType(node, isAType="shape"):
+            # Components are always on shapes. When only a single shape is
+            # parented under a transform Maya returns it as e.g. `cube.f[0:4]`
+            # instead of `cubeShape.f[0:4]` so we expand that to the shape
+            node = cmds.listRelatives(node, shapes=True, fullPath=True)[0]
 
         # Only include valid members of the instance
         if node not in instance_members:


LiborBatek (approved):
I have run it first in current state and it was failing:

Screenshot 2024-09-18 154024

But this PR resolve it and fully functional:


Screenshot 2024-09-18 155803

PR #109, opened 2024-09-15
Validate invisibilities of excluded parents (when "include parent hierarchy" is off)

BigRoy:
## Changelog Description

Add validator for invisibilities on excluded parents (when include parent hierarchy is disabled).

## Additional info

This validation helps to detect the issue where an animator may have hidden or keyed visibilities on parent nodes for an export where these parents are not included in the export. Because if so, those invisibilities would not be included in the export either, giving a different visual result than what the artist likely intended in their workfile.

## Testing notes:

Test both animation and pointcache instances.

1. Create animation or pointcache instance.
2. Set up export with "include parent hierarchy" disabled.
3. On any parent above the objects inside the instance, key the visibility off within the frame range or hide it.
4. The validator should catch it.

So basically for:

```
group/ <- key visibilities or hide it
    cube    <- included in pointcache
```
_with "include parent hierarchy" off, only the cube will be exported; hence the visibility on 'group' is not exported along making the resulting output look different._

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_excluded_parents_visible.py
+++ b/client/ayon_maya/plugins/publish/validate_excluded_parents_visible.py
@@ -1,0 +1,128 @@
+import inspect
+
+from ayon_core.pipeline.publish import (
+    OptionalPyblishPluginMixin,
+    PublishValidationError
+)
+import ayon_maya.api.action
+from ayon_maya.api import plugin, lib
+import pyblish.api
+
+from maya import cmds
+
+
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
It works well when having some visibility keyframes on the root node outside of ```pointcache``` aka not included...

Gives nice validation fail message to the user:
Screenshot 2024-09-24 142826

Once turned on ```Include parent hierarchy``` all good and no validation fail happens..

LGTG!

PR #112, opened 2024-09-16
Validate clashing sibling names when `strip namespaces` is enabled

BigRoy:
## Changelog Description

When exporting siblings that with stripped namespaces share the same name, validate it directly instead of erroring on extraction.

## Additional info

image

Avoids this error on ExtractUSD:
```python
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\dev\ayon-maya\client\ayon_maya\plugins\publish\extract_maya_usd.py", line 357, in process
    cmds.mayaUSDExport(file=file_path,
  File "E:\dev\ayon-maya\client\ayon_maya\plugins\publish\extract_maya_usd.py", line 2, in mayaUSDExport
RuntimeError: Multiple dag nodes map to the same prim path after stripping namespaces: |group1|NewNamespace1:pCube1 - |group1|pCube1
```

Avoids this error on ExtractAlembic:
```python
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\dev\ayon-maya\client\ayon_maya\plugins\publish\extract_pointcache.py", line 227, in process
    extract_alembic(**kwargs)
  File "E:\dev\ayon-maya\client\ayon_maya\api\alembic.py", line 341, in extract_alembic
    cmds.AbcExport(
  File "E:\dev\ayon-maya\client\ayon_maya\plugins\publish\extract_pointcache.py", line 2, in AbcExport
RuntimeError: Alembic Exception encountered: OSchemaObject::OSchemaObject( OObject )
ERROR: EXCEPTION:
Already have an Object named: pCube1
```

## Testing notes:

Create an hierarchy like:

image

1. Export with stripped namespaces enabled.
2. This should be invalidated because without namespaces, both the `pCube1` would share same full path and clash on  export.

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
+++ b/client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
@@ -1,14 +1,15 @@
 from collections import defaultdict
 import inspect
 
-import ayon_maya.api.action
+import pyblish.api
 from ayon_core.pipeline.publish import (
     OptionalPyblishPluginMixin,
     PublishValidationError,
     ValidateContentsOrder,
 )
+
+import ayon_maya.api.action
 from ayon_maya.api import plugin, lib
-import pyblish.api
 
 
 def remove_namespace(path: str) -> str:
--- a/client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
+++ b/client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
@@ -1,0 +1,167 @@
+from collections import defaultdict
+import inspect
+
+import ayon_maya.api.action
+from ayon_core.pipeline.publish import (
+    OptionalPyblishPluginMixin,
+    PublishValidationError,
+    ValidateContentsOrder,
+)
+from ayon_maya.api import plugin, lib
+import pyblish.api
+
+
+def remove_namespace(path: str) -> str:
+    """Remove namespace from full path.
+
+    Example:
+        >>> remove_namespace("|aa:bb:foo|aa:bb:bar|cc:hello|dd:world")
+        '|foo|bar|hello|world'
+
+    Arguments:
+        path (str): Full node path.
+
+    Returns:
+        str: Node path with namespaces removed.
+    """
+    return "|".join(
+        name.rsplit(":", 1)[-1] for name in path.split("|")
+    )
+
+
+class ValidateClashingSiblingNames(plugin.MayaInstancePlugin,
+                                   OptionalPyblishPluginMixin):
+    """Validate siblings have unique names when namespaces are stripped."""
+
+    order = ValidateContentsOrder
+    families = ["pointcache", "animation", "usd"]
+    label = "Validate clashing sibling names"
+    actions = [ayon_maya.api.action.SelectInvalidAction]
+
+    @staticmethod
+    def get_invalid(instance):
+        """Return all nodes that have non-unique names with siblings when
+        namespaces are stripped.
+
+        Returns:
+            list[str]: Non-unique siblings
+        """
+        stripped_name_to_full_path = defaultdict(set)
+        for node in instance:
+            stripped_name = remove_namespace(node)
+            stripped_name_to_full_path[stripped_name].add(node)
+
+        invalid: "list[str]" = []
+        for _stripped_name, nodes in stripped_name_to_full_path.items():
+            if len(nodes) > 1:
+                invalid.extend(nodes)
+
+        if invalid:
+            # We only care about the highest conflicts since child conflicts
+            # only occur due to the conflicts higher up anyway
+            invalid = lib.get_highest_in_hierarchy(invalid)
+
+        return invalid
+
+    def process(self, instance):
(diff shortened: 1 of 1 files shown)


client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
```diff
@@ -0,0 +1,167 @@
...
+
+            report_list = "\n".join(f"- {node}" for node in sorted(invalid))
+
+            raise PublishValidationError(
+                "With stripped namespaces there are conflicting sibling names "
+                "that are not unique:\n"
+                f"{report_list}",
+                description=self.get_description())
+
+    def is_strip_namespaces_enabled(self, instance) -> bool:
```
BigRoy, line 85: @iLLiCiTiT this really seems like madness - but I'm admittedly not entirely sure how to do this better. Many of these plugins can be enabled/disabled, they can be optional and hence disabled for the instance only, they can either have exposed publish attributes for `stripNamespaces` or have it locked down and defined by settings as defaults.

moonyuet:
Have done some simple test with animation product type in my side. 
image

But maybe there is something wrong about the detection of the plugin(it could be my misunderstanding)
`DEBUG: Plugin ExtractAnimation is disabled for this instance.`
I actually enable collect fbx animation and ExtractAnimation should be enabled and run after this.
image

client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
iLLiCiTiT: That actually sounds more legit to me. Single plugin that is capable of showing all the input fields and based on the checked fields adds families (or different data) to instance. But that is a lot of work, so I don't know if should be handled in this PR.

`OptionalPyblishPluginMixin` was not really meant that it is dependent on other plugins. BTW shouldn't the other plugin add something to instance data (for now) instead of having this logic?

BigRoy: > OptionalPyblishPluginMixin was not really meant that it is dependent on other plugins. BTW shouldn't the other plugin add something to instance data (for now) instead of having this logic dependent on current state of instance (might not represent state of instance at the moment the other plugin was processed)?

The other plugins are Extractors - they run after the validators and hence the data is not there yet, so I'd need to move those to a Collector instead - which means we get into the shitshow of backwards compatibilities, etc. :)

This logic here is the best I could think of to support it now without refactoring everything.

BigRoy:
> Have done some simple test with animation product type in my side. image
> 
> But maybe there is something wrong about the detection of the plugin(it could be my misunderstanding) `DEBUG: Plugin ExtractAnimation is disabled for this instance.` I actually enable collect fbx animation and ExtractAnimation should be enabled and run after this. image

@moonyuet Hard to tell from your screenshots whether the `ExtractAnimation` (which is what generates the Alembic; not the FBX) is enabled. Also, the FBX exporter isn't problematic - because looking at the code it only strips a "shared namespace" and not a namespace per object so I'm not sure if that runs into an issue.

So please validate with the Alembic and USD exports - disregard the FBX ones.

LiborBatek (approved):
Works well, been tested on ```pointcache``` product with multiple scenarios (strip namespaces globally on/off, exposed strip namespaces as override etc)

note: there could possibly be super rare edge case with ```model``` outputted also into ```Alembic``` as now the validator doesnt kick in (even when having Strip namespaces ON) but there are other Validators which takes care about ```Namespaces``` within the model content and ```Suffix Naming Conventions``` forbidding names with ":" inside...so this would happen only when those 2 validators disabled...leading into Alembic with non unique naming when stripping on.

client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
```diff
@@ -0,0 +1,167 @@
...
+import inspect
+
+import ayon_maya.api.action
+from ayon_core.pipeline.publish import (
+    OptionalPyblishPluginMixin,
+    PublishValidationError,
+    ValidateContentsOrder,
+)
+from ayon_maya.api import plugin, lib
+import pyblish.api
```
iLLiCiTiT, line 11:
```suggestion
import pyblish.api
from ayon_core.pipeline.publish import (
    OptionalPyblishPluginMixin,
    PublishValidationError,
    ValidateContentsOrder,
)

import ayon_maya.api.action
from ayon_maya.api import plugin, lib
```

client/ayon_maya/plugins/publish/validate_clashing_sibling_names.py
iLLiCiTiT: I do agree "correct implementation" would require massive refactor to allow it. I'll let @antirotor to decide.

moonyuet (approved):

PR #115, opened 2024-09-17
Optimize output data from Maya USD animation export

BigRoy:
## Changelog Description

Reduce amount of data extracted to USD from Maya USD animation.

## Additional info

Draft.

## Testing notes:

1. Publish maya USD for `animation` product - it should work.
2. Check the resulting published data in the USD file, confirm it does not contain materials, UVs, etc.

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_maya_usd.py
+++ b/client/ayon_maya/plugins/publish/extract_maya_usd.py
@@ -240,6 +240,10 @@ class ExtractMayaUsd(plugin.MayaExtractorPlugin,
         # Can be overridden by inherited classes
         return members
 
+    def override_options(self, options: dict) -> dict:
+        # Can be overridden by inherited classes
+        return options
+
     def process(self, instance):
         if not self.is_active(instance.data):
             return
@@ -343,6 +347,8 @@ class ExtractMayaUsd(plugin.MayaExtractorPlugin,
                 )
                 del options[key]
 
+        options = self.override_options(instance)
+
         self.log.debug('Exporting USD: {} / {}'.format(file_path, members))
         with maintained_time():
             with maintained_selection():
@@ -416,6 +422,14 @@ class ExtractMayaUsdAnim(ExtractMayaUsd):
 
         members = cmds.ls(cmds.sets(out_set, query=True), long=True)
         return members
+
+    def override_options(self, options: dict) -> dict:
+        options = options.copy()
+        options["exportUVs"] = False
+        options["exportColorSets"] = False
+        options["exportMaterials"] = False
+        options["exportAssignedMaterials"] = False
+        return options
 
 
 class ExtractMayaUsdModel(ExtractMayaUsd):


client/ayon_maya/plugins/publish/extract_maya_usd.py
```diff
@@ -343,6 +347,8 @@ def parse_attr_str(attr_str):
                 )
                 del options[key]
 
+        options = self.override_options(instance)
```
moonyuet, line 350: Is it supposed that the argument should be dictionary?
It errors out the extractor as shown below:
```
DEBUG: Export options: {'defaultUSDFormat': 'usdc', 'stripNamespaces': False, 'mergeTransformAndShape': True, 'exportDisplayColor': False, 'exportColorSets': True, 'exportInstances': True, 'exportUVs': True, 'exportVisibility': True, 'exportComponentTags': False, 'exportRefsAsInstanceable': False, 'eulerFilter': True, 'renderableOnly': False, 'shadingMode': 'none', 'convertMaterialsTo': 'none', 'jobContext': None, 'filterTypes': None, 'staticSingleSample': True, 'worldspace': False}
DEBUG: Performing extraction ...
DEBUG: Collected objects: ['|asset_rigMain_01_:_GRP', 'asset_rigMain_01_:rigMain_out_SET', 'asset_rigMain_01_:rigMain_controls_SET', 'asset_rigMain_01_:rigMain_skeletonMesh_SET', 'asset_rigMain_01_:rigMain_skeletonAnim_SET']
DEBUG: options: <class 'list'>
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\dependency_packages\ayon_2402141620_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\ayon-addon_template\ayon-maya\client\ayon_maya\plugins\publish\extract_maya_usd.py", line 350, in process
    options = self.override_options(instance)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\ayon-addon_template\ayon-maya\client\ayon_maya\plugins\publish\extract_maya_usd.py", line 429, in override_options
    options["exportUVs"] = False
    ~~~~~~~^^^^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
```
[publish-report-241023-19-04.json](https://github.com/user-attachments/files/17489825/publish-report-241023-19-04.json)


As it is overriding the option, shouldn't it be: 
```suggestion
        options = self.override_options(options)
```

client/ayon_maya/plugins/publish/extract_maya_usd.py
```diff
@@ -417,6 +423,14 @@ def filter_members(self, members):
         members = cmds.ls(cmds.sets(out_set, query=True), long=True)
         return members
 
+    def override_options(self, options: dict) -> dict:
+        options = options.copy()
+        options["exportUVs"] = False
+        options["exportColorSets"] = False
+        options["exportMaterials"] = False
+        options["exportAssignedMaterials"] = False
```
moonyuet, line 431: I tested out with these two options and also do some research on the [mayaUSD](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/commands/Readme.md). It seems that these two options haven't been exposed to Python as it errors out in maya 2025.1 during extraction.
image

image

BigRoy: That's odd. But they are in the list here: https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/commands/Readme.md#mayausdexportcommand

Can you check which `maya-usd` addon version you're using?

moonyuet: I am at 0.28.0 not the the latest 0.30.0. That could be the reason

BigRoy: Thanks - yes. Seems added in 0.29.0 -> https://github.com/Autodesk/maya-usd/blob/v0.29.0/doc/CHANGELOG.md

Will take it into consideration. 👍

moonyuet:
I tried with the latest version of maya 2025.3 with mayausd 0.30.0
The animation is published successfully.
image

@BigRoy Do we need to implement some backward compatibility for just in case users are using maya-usd 0.28.0 or lower?

BigRoy:
> I tried with the latest version of maya 2025.3 with mayausd 0.30.0 The animation is published successfully. image
> 
> @BigRoy Do we need to implement some backward compatibility for just in case users are using maya-usd 0.28.0 or lower?

Yes, that's the idea. Do keep in mind this PR is still in draft. There are other things to solve too, like:

- This may need to be optional or configurable in settings. (especially for backwards compatibility with existing projects)
- Shouldn't crash on older USD addons, but if the setting is used on the instance what should it do then? Log a warning that the output might not consider that and hence may be different than what the user expects from the toggles?

moonyuet:
> > I tried with the latest version of maya 2025.3 with mayausd 0.30.0 The animation is published successfully. image
> > @BigRoy Do we need to implement some backward compatibility for just in case users are using maya-usd 0.28.0 or lower?
> 
> Yes, that's the idea. Do keep in mind this PR is still in draft. There are other things to solve too, like:
> 
> * This may need to be optional or configurable in settings. (especially for backwards compatibility with existing projects)

> * Shouldn't crash on older USD addons, but if the setting is used on the instance what should it do then? Log a warning that the output might not consider that and hence may be different than what the user expects from the toggles?

This should be configurable in settings, but we need to inform users about the missing functionalities if they are using elder versions of maya usd.

PR #118, opened 2024-09-19
Fix validate unit settings

BigRoy:
## Changelog Description

Fixes the validate units configuration when any setting for it was overridden.

## Additional info

Apparently the labels and values were reversed.
Came up on discord [here](https://discord.com/channels/517362899170230292/1286380990171713606)

## Testing notes:

1. Create addon
2. Upload to bundle
3. Set custom validate unit settings
4. It should validate accordingly.

### DIFF
--- a/server/settings/publishers.py
+++ b/server/settings/publishers.py
@@ -16,22 +16,22 @@ from .publish_playblast import (
 def linear_unit_enum():
     """Get linear units enumerator."""
     return [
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Has been tested with few different settings (mm, centimeters, inches) and also matching / non matching maya scene units set...all works well.

Didnt spot any issues.

Screenshot 2024-09-20 135927

PR #120, opened 2024-09-25
Layout: Do not error on extraction with renamed instance node

BigRoy:
## Changelog Description

Do not error if `instance` does not match the actual node's name + re-use `setMembers` which was already collected anyway.

## Additional info

Quick fix.

## Testing notes:

1. Extract Layout should work as intended.


LiborBatek (approved):
It does what it suppose to do...LGTM

before this PR

Screenshot 2024-09-25 135522

and the same source workfile using this branch:



Screenshot 2024-09-25 135957

PR #126, opened 2024-09-26
Skip extraction of active view for automatic tests

kalisp:
## Changelog Description
It seems that Maya UI is not completely visible or shutting down, `view.readColorBuffer` causes RuntimeError: (kFailure): Unexpected Internal Failure aas view is not visible.

## Additional info
This applies only to running Maya in automatic tests.

## Testing notes:

1. publish in Maya ordinarily, active view thumbnail should be still captured for workfile instance


LiborBatek (approved):
I have got extracted thumb for workfile without any issues, LGTM

image

PR #127, opened 2024-09-26
Make sure extract layout does not publish empty json

moonyuet:
## Changelog Description
This PR is to make sure extract layout does not publish empty json

## Additional info
Please try testing when groupLoadAssets option enabled 

## Testing notes:

1. Create Layout
2. Extract Layout


moonyuet:
the layout does not work because of other issue which is resolved by #104

LiborBatek (approved):
All seems fine and working...I was able to produce succesful publish (checked that JSON been correct) and also tried to load ```layout``` product back into maya...all good!

Screenshot 2024-09-30 183504

once loaded back to maya:

Screenshot 2024-09-30 183908

PR #132, opened 2024-10-01
Maya USD rigs with asset and shot contributions

### LINKED ISSUE #10: Enhancement: Maya USD: Rig in USD Asset as MayaReference prim

Working within a USD Pipeline usually means you want your animation output for e.g. a rig to be overlaid on top of a loaded USD Asset to merge the animated point data with the model and look of the asset, usually for use by downstream lighting department.

The USD reference usually provides:
```
- asset (referenced usd asset)
  - geo
  - mtl
```

The output data from a rig should then be overlaid over the `/asset/geo` path in the shot to move the character around.

Currently, publishing an animation from a rig usually means completely separate Alembic or USD data that would need to _somehow_ be identified to get overlaid over a particular USD asset structure in a shot.

What if we could load an asset, and it itself, has a reference to the rig embedded?

Maya USD allows to embed maya scene files (mayaAscii or mayaBinary) to be included into USD as a `MayaReference` prim referencing the file. When a Maya USD Proxy finds such a prim it can enable the prim as a regular maya reference to load the native maya data, like for example a rig file.

I'd like to propose to - whenever publishing a **rig** from Maya to write out a USD Asset Contribution where an extra layer gets  added to the USD asset representing the rig as a reference.

Basically someone (e.g. either in Maya or Houdini) lays out assets in a scene, by loading the asset USD. If that asset now contains the `MayaReference` prim (which can be loaded fine in Houdini; just the internal data of the maya scene couldn't be loaded) then in Maya one could now enable that reference inside the USD Asset to load the actual rig.

So assets laid out in any USD supporting host could in Maya be switched to their rig for animating. Nice!

In essence we'd need:

1. On publishing a Maya rig, create a USD asset contribution that references in this rig version in a USD layer
2. On Maya, when the Maya USD reference rig is loaded (and its detected its loaded within USD asset) then publish the animation cache as USD that gets added as a variant on the loaded asset in the shot's anim layer.

That way, the animator in Maya can just read a USD Layout with loaded assets, enable the reference, animate and publish.

Plus side to this would be, that when they've published their animation to the shot's layers they can also disable their reference and switch it over it to the published cache (since it's also a USD contribution). Meaning they can use the cached USD data as proxy in their scene whilst maybe e.g. continuing to animate on another character.

Loading rigs manually, then publishing and manually overlaying (e.g. referencing) the output over the USD data over an asset, after the fact.

### Additional context:

_No response_

[cuID:[OP-7978](https://app.clickup.com/t/86bx4bjau)]

BigRoy:
## Changelog Description

Add support to publish maya rig directly into USD asset, animate it and directly publish into the USD shot

## Additional info

Fixes #10 

## Testing notes:

_TODO_

1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_maya/plugins/create/create_animation_pointcache.py
+++ b/client/ayon_maya/plugins/create/create_animation_pointcache.py
@@ -1,11 +1,19 @@
+from collections import namedtuple
+from typing import Optional, TYPE_CHECKING
+
 from maya import cmds
 
+from ayon_core.pipeline.create import CreatedInstance
 from ayon_maya.api import lib, plugin
 
 from ayon_core.lib import (
     BoolDef,
     NumberDef,
 )
+
+if TYPE_CHECKING:
+    from pxr import Usd
+    import ufe.PyUfe
 
 
 def _get_animation_attr_defs(create_context):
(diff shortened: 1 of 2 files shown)


client/ayon_maya/plugins/create/create_animation_pointcache.py
```diff
@@ -132,3 +140,203 @@ def create(self, product_name, instance_data, pre_create_data):
...
+            # No maya usd plug-in
+            return
+
+        associated_nodes = cmds.listConnections(
+            f"{reference_node}.associatedNode")
+        if not associated_nodes:
+            return
+
+        for node in associated_nodes:
+            if mayaUsdUtils.isPulledMayaReference(node):
```
moonyuet, line 325:
```suggestion
            if mayaUsdUtils.isPulledMayaReference(node):
```
I strangely hit this error when I tested with maya 2025.1.
image

BigRoy: Thanks. I did notice some issues in Maya 2025 with regards to their USD API too.

I also saw this in their recent [release notes of Maya 2025.3](https://help.autodesk.com/view/MAYAUL/2025/ENU/?guid=GUID-59686914-791F-4386-8B0C-78DBEA96B029):

image

Which is also something I'll need to double check.

PR #133, opened 2024-10-01
Update handle start and handle end on instances on task change

### LINKED ISSUE #76: Update instance frame range include handles

With the fixed behavior for Updating instance frame range when saving to another context with PR https://github.com/ynput/ayon-maya/pull/38 it now also becomes relevant to discuss what to do with Handles.

Currently, the update/reset logic does not set `handles` at all for the instances.

We should maybe include by default also the handles on resetting.

Side note: Maybe this should be tweakable via settings. Initially I thought think we're better of forcing the default behavior to include maybe the handles. However, by default maya instances set their frame ranges according to the time slider - not based of of the task entity's current settings. As such, it suddenly resetting to folder on switch is already confusing - and maybe even more confusing that it suddenly includes handles?

_No response_

### Additional context:

This issue basically is coming from [this comment](https://github.com/ynput/ayon-maya/pull/38#issuecomment-2248847990):

> ### Question 1: Should it also update frame ranges?
> 
> The code does not update the handle start or handle end. And I wonder, should it? If so, we'd need to add:
> 
> ```python
>         "handleStart": float(task_entity["attrib"]["handleStart"]),
>         "handleEnd": float(task_entity["attrib"]["handleEnd"]),
> ```
> 
> ### Question 2: Should updating of the frame ranges become a separate checkbox from updating the target context (folder/task) in the instance?
> 
> Would there be cases where you do want to update the instance to the new folder+task but do not want to update the frame ranges for the instances to the new context? If so, should it be a separate checkbox in the dialog?

BigRoy:
## Changelog Description

Update handle start and handle end on instances on task change when updating to match the context

## Additional info

Fixes #76 

## Testing notes:

1. Set up workfile with publish instances that have handle start and handle end, e.g. pointcache instances.
2. Save workfile to different context.
3. A popup dialog shows, asking to update instances, etc. Apply it.
3. The instance's frame range should for handle start and handle end should now be updated to that of the new task entity.

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -3243,6 +3243,8 @@ def update_content_on_context_change():
     creator_attribute_values = {
         "frameStart": float(task_entity["attrib"]["frameStart"]),
         "frameEnd": float(task_entity["attrib"]["frameEnd"]),
+        "handleStart": float(task_entity["attrib"]["handleStart"]),
+        "handleEnd": float(task_entity["attrib"]["handleEnd"]),
     }
 
     has_changes = False


moonyuet (approved):
Didn't test, code looks good for me.

LiborBatek (approved):
Working well...no any issues spotted, looks ok!

Screenshot 2024-10-02 121744

After changing the context also Publish instances got updated (handles + frame start/end)

PR #148, opened 2024-10-11
Fix including of containers with e.g. `layout` products

BigRoy:
## Changelog Description

With this fix the `layout` product now contains the `containers` of the exported assets again. So that, upon loading, the assets inside the layout also appear in the scene inventory.

## Additional info

This being broken was a regression - so this restores the original behavior.

## Testing notes:

1. Publishing and loading the layout again should preserve the containers


moonyuet (approved):
Code looks good for me. And did the simple test. The scene inventory does not go empty anymore.
image

LiborBatek (approved):
Looks working fine...

I had to produce new publish of ```layout``` product to make it work (not working with older published products)

but then all good , here is my source scene for publishing with loaded assets:

Screenshot 2024-10-14 172454

and ```Scene Inventory``` when loaded ```layout``` product in the workfile


Screenshot 2024-10-14 172700

PR #152, opened 2024-10-16
Maya USD Proxy Loader: Do not use AYON Entity URI by default

BigRoy:
## Changelog Description

Disable using the AYON Entity URI by default for Load Maya USD proxy loader.

## Additional info

It should be up to the studio to enable it if they are using the asset resolver.

## Testing notes:

1. Using default settings loading with Maya USD Proxy Loader it should not use the AYON Entity URI.


moonyuet (approved):
LGTM

LiborBatek (approved):
Once loaded the ```USD``` asset as proxy via ```Load USD (Proxy)``` it correctly uses standard file path:
Screenshot 2024-10-16 135100

PR #163, opened 2024-10-24
Create plugins: Apply settings correctly

iLLiCiTiT:
## Changelog Description
Create Render and Vray Scene are propagating settings.

## Additional info
The plugins did not propagate their settings because were collecting only `render_settings`. Also change the classmethods to methods.

## Testing notes:
1. Change of `enabled` state in settings on both create plugins should be propagated.

Resolves https://github.com/ynput/ayon-maya/issues/162

### DIFF
--- a/client/ayon_maya/api/plugin.py
+++ b/client/ayon_maya/api/plugin.py
@@ -1,10 +1,8 @@
 import json
 import os
-from abc import ABCMeta
 
 import ayon_api
 import qargparse
-import six
 
 from ayon_core.lib import BoolDef, Logger
 from ayon_core.pipeline import (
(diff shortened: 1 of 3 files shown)


LiborBatek (changes requested):
It does work and propagate settings from addon settings but just for

```VrayProxy``` but for some reason not working for ```VrayScene```...
Screenshot 2024-10-24 154317

iLLiCiTiT:
> VrayProxy but for some reason not working for VrayScene...

What is not working? I changed behavior related to `enabled` state and default variants. Nothing else.

LiborBatek:
> > VrayProxy but for some reason not working for VrayScene...
> 
> What is not working? I changed behavior related to `enabled` state and default variants. Nothing else.

I get that...so to be specific, ```Variant``` if set some new one besides ```Main``` is not visible when testing ```VRayScene```....it all works for both enabled and variants when speaking of ```VRayProxy```...

PR #164, opened 2024-10-24
Adds skeletal mesh extractors as configurable settings

Kryses:
## Changelog Description
Adds options in the extractor settings to turn on or off the skeletal mesh abc and fbx extractors.

image

## Additional info
At Halon we don't typically use the Abc export for skeletal meshes and needed a way to disable to extractor. One thing to note about the ABC extractor is that either we're using it wrong or it has some issues with it. 

image

## Testing notes:

1. Open a rig working file
2. In Create add a new `Unreal - Skeletal Mesh` publish instance.
3. Make sure to add your mesh and you joints to the  `geometry_SET` and the `joints_set`
4. Depending on your server settings you should see the following on the publish instance.
image
5. Click publish.

You should see that it only exports what ever options you have selected.

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_unreal_skeletalmesh_abc.py
+++ b/client/ayon_maya/plugins/publish/extract_unreal_skeletalmesh_abc.py
@@ -5,10 +5,11 @@ import os
 from ayon_maya.api.alembic import extract_alembic
 from ayon_maya.api.lib import maintained_selection, suspended_refresh
 from ayon_maya.api import plugin
+from ayon_core.pipeline import publish
 from maya import cmds  # noqa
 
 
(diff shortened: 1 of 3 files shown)


moonyuet:
Thank you for your contribution. Speaking of the rig workflow. there is already an existing fbx rig/animation workflow for publishing the rig to unreal. You need to put your character geometry into the `skeletonMesh_SET` and it would export the rig in fbx for you.
You can find this community post helpful for you: https://community.ynput.io/t/loading-looks-in-rigging-scenes/1892/11

Kryses:
ah perfect thanks for pointing that out. Just did a quick run through of that and it's doing the trick. I'll go ahead and close this out.

PR #167, opened 2024-10-29
Support assigning maya look to multiple separate assets inside an Arnold `aiStandin` with alembic files.

### LINKED ISSUE #160: AY-7024_Look Assigner to combined aiStanInd issue
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

So, take 2 assets that have been published as models (with abc extractor included) and have a look also published. Load the abc representation of those 2 assets, create a pointche instance and publish an alembic with those 2 assets together.  Let's call this asset, a_b_combined If you load a_b_combined as a referenced alembic, you can run the look assigner with no issue, both assets in the combined asset are detected, looks related as well and all works fine. Now, if you load a_b_combined as a standIn, the look manager detects both assets and it's relative look, BUT the asignment will go wrong.  No errors, just the logic fails since the name of the container of the alembic procedural will be the same for both assets.  I think here is the logic key to this. This creates 2 issues: 
even if you only select object a and assing it's look, all looks of all asset in the combined asset will be loaded
the assignemnt of looks will always be of the last look to the last asset in the loop since all nodes in the function will have the same name.

### How would you imagine the implementation of the enhancemenent?
_No response_

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1298936279924215869/1298936283565129748)

_This issue was automatically created from Clickup ticket [AY-7024](https://app.clickup.com/t/86c0t8vku)_

BigRoy:
## Changelog Description

1. Do not force remove assignments on aiStandin look assignments, but allow updating/adding assingments so that assigning looks of multiple assets inside one `aiStandin` shape is supported.

2. This also fixes a bug where `cbId` attribute exported from Houdini was not detected correctly and its value was incorrectly interpreted from the alembic property.
- Maya seems to export as `alembic.Abc.IArrayProperty`
- Houdini seems to export as `alembic.Abc.IScalarProperty`

3. You can now also list assets and assign look to `gpuCache` if `mtoa` (Arnold) renderer is loaded and is set as the current renderer.

4. This also fixes a bug where ALL assets inside an `aiStandin` are assigned even if only one of them was picked from the asset list in the look assigner.

We now support both.

## Additional info

Fixes #160 

## Testing notes:

1. Create a pointcache containing geometry from multiple folders (assets)
2. Load the Alembic pointcache via AYON to `aiStandin`.
3. Assign the look.

It should detect the multiple assets inside the `aiStandin` and allow assigning each of them without applying only the last asset.

### DIFF
--- a/client/ayon_maya/tools/mayalookassigner/alembic.py
+++ b/client/ayon_maya/tools/mayalookassigner/alembic.py
@@ -71,8 +71,14 @@ def get_alembic_paths_by_property(filename, attr, verbose=False):
         if not _property.isConstant():
             log.warning("Id not constant on: {0}".format(name))
 
-        # Get first value sample
-        value = _property.getValue()[0]
+        # Maya alembic export seems to export `cbId` string as array property
+        # whereas Houdini primitive `cbId` attribute comes through as a
+        # scalar property (which still holds the string?)
+        if isinstance(_property, alembic.Abc.IArrayProperty):
+            # Get first value sample
+            value = _property.getValue()[0]
+        else:
+            value = _property.getValue()
 
         obj_ids[name] = value
 
--- a/client/ayon_maya/tools/mayalookassigner/app.py
+++ b/client/ayon_maya/tools/mayalookassigner/app.py
@@ -268,7 +268,14 @@ class MayaLookAssignerWindow(QtWidgets.QWidget):
 
             # Assign Arnold Standin look.
             if cmds.pluginInfo("mtoa", query=True, loaded=True):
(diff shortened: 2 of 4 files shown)


moonyuet (approved):
Tested successfully in Maya: 
image

LiborBatek (approved):
I have used 3 different assets for ```pointcache``` and loaded it as a ```standin``` and using ```Look assigner``` correctly resulting into all looks applied to each entity inside! Really cool!

LGTM

Screenshot 2024-10-30 115249

And after using assign look:

Screenshot 2024-10-30 115403

Issue #168, opened 2024-11-04
AY-7099 Publishing rig without out_SET should be disallowed or loading it should be allowed.


BigRoy:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior

It seems that the Validate Rig Contents [is disabled by default](https://github.com/ynput/ayon-maya/blob/fb91a6d4bbe94720efa154164ba237ea7848bfb3/server/settings/publishers.py#L1454-L1458) however that plug-in is the plug-in that enforces the `out_SET` of a rig [to exist and have members](https://github.com/ynput/ayon-maya/blob/06aa2445b4df511225b6795b283dd974040c0b5f/client/ayon_maya/plugins/publish/validate_rig_contents.py#L49-L53).

Without that `rig_SET` loading the rig fails:

![Image](https://github.com/user-attachments/assets/e5f3b0d2-1f65-4866-bd41-04899915c26a)

## Expected Behavior

1. Either that `out_SET` must be enforced and that validator can not be disabled or made optional.
2. Or loading the rig without `out_SET` should be allowed without errors.

### Steps To Reproduce:

1. Create rig instance
2. Delete `out_SET` inside the instance
3. Publish
4. Load the rig product

PR #169, opened 2024-11-04
Allow loading published rig that has no `out_SET` and no `controls_SET`

BigRoy:
## Changelog Description

Allow loading published rig that has no `out_SET` and no `controls_SET` but then skip creating the animation instance automatically, with a warning

There may be edge cases where a rig was published without those sets without validation errors (the validators are disabled by default) but then being unable to load them is very annoying. So this allows them to be loaded, but with a warning logged instead.

Fixes https://github.com/ynput/ayon-maya/issues/168

## Testing notes:

1. Create rig instance
2. Remove the `out_SET` and `controls_SET` from the instance
3. (Make sure the validate rig contents validator is disabled; it's disabled by default)
4. Publish rig
5. Load the rig

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -83,6 +83,15 @@ DISPLAY_LIGHTS_ENUM = [
 ]
 
 
+class RigSetsNotExistError(RuntimeError):
+    """Raised when required rig sets for animation instance are missing.
+
+    This is raised from `create_rig_animation_instance` when the required
+    `out_SET` or `controls_SET` are missing.
+    """
+    pass
+
+
 def get_main_window():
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
I have followed the testing steps and all been functional...

ended up with loaded rig without ```out_set``` and ```control_set``` without any issues and no ```animation``` publish instance present within publisher ui.

image

BigRoy:
> I have followed the testing steps and all been functional...
> 
> ended up with loaded rig without `out_set` and `control_set` without any issues and no `animation` publish instance present within publisher ui.
> 
> image

Just to be sure - was there a warning logged in the script editor?

PR #170, opened 2024-11-05
Extract Layout: Do not error on containers loaded from library project

BigRoy:
## Changelog Description

Extract Layout: Do not error on containers loaded from library project - instead log a warning and skip them.

This may fix the issue reported here: https://community.ynput.io/t/manage-tab-bug-when-using-library-project-assets/1612/9?u=bigroy

## Testing notes:

1. Extract layout should work but with warnings if products are included that are not loaded from current project
2. It should now also ignore safely containers that have invalid uuids, e.g. assets still originating from OpenPype or containers that for whatever reason have a broken `representation` id attribute value.


LiborBatek (approved):
I was able to produce a ```layout``` which contained the asset from ```library``` project and without any errors just warning within publish report.

Looks good to me...

Screenshot 2024-11-05 102741

PR #172, opened 2024-11-08
USD Export: Workaround default prim bug in Maya 2025.3 with Maya USD 0.30.0

BigRoy:
## Changelog Description

Workaround default prim bug in Maya 2025.3 with Maya USD 0.30.0

See: https://github.com/Autodesk/maya-usd/issues/3991

Without this fix the published USD file will not be loaded correctly when referenced. This is e.g. noticable when using the AYON USD contribution workflow and the model is 'missing' when loading the asset USD file. This is because the generated file itself lacks correct default prim metadata due to the Maya USD regression/bug.

## Testing notes:

1. Publish a Maya USD file from different versions of Maya (to confirm they all still work).
2. The generated USD asset files (with contribution workflow) should work.

Or simpler, but maybe more technical for some. Confirm the actual written `defaultPrim` value in the written USD file is not prefixed with a `|`.

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_maya_usd.py
+++ b/client/ayon_maya/plugins/publish/extract_maya_usd.py
@@ -343,6 +343,23 @@ class ExtractMayaUsd(plugin.MayaExtractorPlugin,
                 )
                 del options[key]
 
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Didnt tested but LGTM...

PR #176, opened 2024-11-12
Load: Containerize `project_name` on load to allow managing containers from different projects

BigRoy:
## Changelog Description

Maya imprint and update with `project_name` on `container` object.

Should be tested together with PR https://github.com/ynput/ayon-core/pull/1005

**Note** that this would only set project name on update for `ReferenceLoader` plug-ins and not the other, so that would still be TODO. However, that'd only be relevant when opening a scene from another project OR when being able to 'switch asset' to another project.

## Testing notes:
1. Use with PR https://github.com/ynput/ayon-core/pull/1005
2. Load containers from library project into current project.
3. They should be manageable via the scene inventory.

OR alternatively:

1. Load containers from current project.
2. Save workfile.
3. Launch maya from another project - secretly open the file from the first project.
4. It should list and manage the containers just fine in the scene inventory because it now knows about its project.

### DIFF
--- a/client/ayon_maya/api/plugin.py
+++ b/client/ayon_maya/api/plugin.py
@@ -947,7 +947,7 @@ class ReferenceLoader(Loader):
 
         # Update metadata
         for attr_name, value in {
-            "representation": repre_entity["_id"],
+            "representation": repre_entity["id"],
             "project_name": context["project"]["name"]
         }.items():
             lib.set_attribute(node=node, attribute=attr_name, value=value)
--- a/client/ayon_maya/api/plugin.py
+++ b/client/ayon_maya/api/plugin.py
@@ -946,9 +946,11 @@ class ReferenceLoader(Loader):
             cmds.sets(invalid, remove=node)
 
         # Update metadata
-        cmds.setAttr("{}.representation".format(node),
-                     repre_entity["id"],
-                     type="string")
+        for attr_name, value in {
+            "representation": repre_entity["_id"],
+            "project_name": context["project"]["name"]
+        }.items():
+            lib.set_attribute(node=node, attribute=attr_name, value=value)
 
         # When an animation or pointcache gets connected to an Xgen container,
         # the compound attribute "xgenContainers" gets created. When animation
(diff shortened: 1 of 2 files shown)


client/ayon_maya/api/plugin.py
```diff
@@ -946,9 +946,11 @@ def update(self, container, context):
             cmds.sets(invalid, remove=node)
 
         # Update metadata
-        cmds.setAttr("{}.representation".format(node),
-                     repre_entity["id"],
-                     type="string")
+        for attr_name, value in {
+            "representation": repre_entity["_id"],
```
iLLiCiTiT, line 950:
```suggestion
            "representation": repre_entity["id"],
```

client/ayon_maya/api/plugin.py
```diff
@@ -946,9 +946,11 @@ def update(self, container, context):
...
+            "project_name": context["project"]["name"]
+        }.items():
```
iLLiCiTiT, line 952:
```suggestion
        for attr_name, value in [
            ("representation", repre_entity["_id"]),
            ("project_name", context["project"]["name"]),
        ]:
```

client/ayon_maya/api/plugin.py
```diff
@@ -946,9 +946,11 @@ def update(self, container, context):
             cmds.sets(invalid, remove=node)
 
         # Update metadata
-        cmds.setAttr("{}.representation".format(node),
-                     repre_entity["id"],
-                     type="string")
+        for attr_name, value in {
+            "representation": repre_entity["id"],
+            "project_name": context["project"]["name"]
+        }.items():
```
BigRoy, line 952:
```suggestion
        for attr_name, value in [
            ("representation", repre_entity["id"]),
            ("project_name", context["project"]["name"]),
        ]:
```

client/ayon_maya/api/plugin.py
BigRoy: Haha thanks - my IDE with copilot installed is starting to have a mind of its own sometimes.

iLLiCiTiT (approved):
I'd say this needs validation of @antirotor . But if we do want to support per project loading, then, yes?

PR #178, opened 2024-11-14
Avoid being able to undo a part of containerisation

BigRoy:
## Changelog Description

Containerise will now happen inside an undo chunk, disallowing to undo a part of the command.

Without this there is a bug if the user were to Load something and they would press UNDO a few times that they may have undo'ed a part of the containerisation. In particular, they may have undo'ed to JUST a state where the container is recognized but it lacks the representation attribute - meaning it will break anything that tries to get the `container["representation"]` value from the containers in the scene - e.g. loader, publishing from the scene (on collect input versions), etc.

This makes it so that if the user were to undo - it undoes the full creation of the container.

Technically we could ALSO add the same undo chunk decorator on all the maya loader plug-in methods to ensure they all operate in a single undo chunk.

## Testing notes:

1. Loading in Maya should work.
2. Undoing should not undo 'parts of the container' like e.g. the creation of attributes on it but the full creation of it in one go.

TIP: To see containers in the outliner enable "Ignore hidden in outliner":
image

### DIFF
--- a/client/ayon_maya/api/pipeline.py
+++ b/client/ayon_maya/api/pipeline.py
@@ -419,6 +419,7 @@ def ls():
         yield parse_container(container)
 
 
+@lib.undo_chunk()
 def containerise(name,
                  namespace,
                  nodes,


LiborBatek (changes requested):
I have made tests with loading of

```model```
```pointcache```
```camera```
```layout```

and all was working well when hitting ```Undo``` and the ```AVALON_CONTAINERS``` was diminshing and gone.

However this doesnt work for ```rig``` products as when loaded and hitting ```Undo``` do nothing and keep the container intact and present...

BigRoy:
> I have made tests with loading of
> 
> `model` `pointcache` `camera` `layout`
> 
> and all was working well when hitting `Undo` and the `AVALON_CONTAINERS` was diminshing and gone.
> 
> However this doesnt work for `rig` products as when loaded and hitting `Undo` do nothing and keep the container intact and present...

Having it remain completely intact and not destroy the scene is **fine** behavior even if maybe unexpected. In this case it's most likely due to the referencing of the rig being part of the creation and you can't undo referencing in Maya.

So if nothing breaks, I think this is actually correct behavior @LiborBatek ?

LiborBatek (approved):
reevaluated after @BigRoy comment and is eligible for approval atm

PR #179, opened 2024-11-14
Skip invalid containers with a warning

BigRoy:
## Changelog Description

Skip invalid containers with a warning - instead of breaking publishi…ng, loader, inventory, etc.

This avoids similar issues as #178 does - but is a 'fix' that would avoid the issue being problematic in existing scenes (instead of just on containerise) or if the container breaks in any other way.

The most likely case for these containers to become invalid was the issue #178 solves by 'bundling' undo so that users can't undo "half of the container creation" deleting only a few of its attributes. However, users can still break things in magical ways, time has shown us.

It would now log e.g. this if the container is incomplete and incomplete containers are not listed in the scene inventory either, because well they are incomplete:

```
// Warning: ayon_maya : Container 'char_hero_pointcacheMain_01__pointcacheMain_CON' is missing required keys: ['loader', 'representation']
```

## Testing notes:

1. Load a few products
2. In outliner, enable: Display > Ignore "Hidden in outliner"
3. Select the containers
4. Randomly delete attributes on these instances, like `loader`, `representation`, etc.

A warning should be logged if any of the required keys ends up missing. The required keys are considered to be `["id", "name", "namespace", "loader", "representation"]`

### DIFF
--- a/client/ayon_maya/api/pipeline.py
+++ b/client/ayon_maya/api/pipeline.py
@@ -347,10 +347,17 @@ def parse_container(container):
         container (str): A container node name.
 
     Returns:
-        dict: The container schema data for this container node.
+        dict[str, Any]: The container schema data for this container node.
 
     """
     data = lib.read(container)
+
+    required = ["id", "name", "namespace", "loader", "representation"]
+    missing = [key for key in required if key not in data]
+    if missing:
+        log.warning("Container '%s' is missing required keys: %s",
+                    container, missing)
+        return {}
 
     # Backwards compatibility pre-schemas for containers
     data["schema"] = data.get("schema", "openpype:container-1.0")
@@ -411,12 +418,14 @@ def ls():
     they are called 'containers'
 
     Yields:
-        dict: container
+        dict[str, Any]: container
 
     """
     container_names = _ls()
     for container in sorted(container_names):
-        yield parse_container(container)
+        container_data = parse_container(container)
+        if container_data:
+            yield container_data
 
 
 def containerise(name,
--- a/client/ayon_maya/api/pipeline.py
+++ b/client/ayon_maya/api/pipeline.py
@@ -368,7 +368,8 @@ def parse_container(container):
         container (str): A container node name.
 
     Returns:
-        dict[str, Any]: The container schema data for this container node.
+        Optional[dict[str, Any]]: The container schema data for this container
+         if it meets all the required metadata.
 
     """
     data = lib.read(container)
@@ -378,7 +379,7 @@ def parse_container(container):
     if missing:
         log.warning("Container '%s' is missing required keys: %s",
                     container, missing)
-        return {}
+        return
 
     # Backwards compatibility pre-schemas for containers
     data["schema"] = data.get("schema", "openpype:container-1.0")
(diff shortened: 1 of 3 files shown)


client/ayon_maya/api/pipeline.py
```diff
@@ -347,11 +347,18 @@ def parse_container(container):
...
 
     Returns:
-        dict: The container schema data for this container node.
+        dict[str, Any]: The container schema data for this container node.
 
     """
     data = lib.read(container)
 
+    required = ["id", "name", "namespace", "loader", "representation"]
+    missing = [key for key in required if key not in data]
```
iLLiCiTiT, line 356:
```suggestion
    required = {"id", "name", "namespace", "loader", "representation"}
    missing = required - set(data.keys())
```

client/ayon_maya/api/pipeline.py
```diff
@@ -347,11 +347,18 @@ def parse_container(container):
...
+    if missing:
+        log.warning("Container '%s' is missing required keys: %s",
+                    container, missing)
+        return {}
```
iLLiCiTiT, line 360: Shoulnd't this return `None` rather than empty dictionary?

BigRoy: Not according to the functions type hint :D
Checked the usage of the function and noticed this may also influence `setdress` in maya - so I may want to do this differently.

LiborBatek (approved):
Works well and when some attributes missing, it still publishes ok and just gives warning about some missing keys...LGTM

image

client/ayon_maya/api/pipeline.py
iLLiCiTiT: Then we should change type hint? I meant returning empty dictionary if there are missing keys is worse than returning whatever is stored in scene.

BigRoy: Implemented with https://github.com/ynput/ayon-maya/pull/179/commits/5d689b57cf45fda4d89490d41ca6fa1c8250b482

LiborBatek:
@BigRoy should I also re-test it?

BigRoy:
> @BigRoy should I also re-test it?

Sorry, missed this. Yes, please.

LiborBatek (commented):
It works when publishing ```animation``` and some attribs are missing on the loaded ```rig``` and gives just warning during the publishing process...

Screenshot 2025-01-29 133802

but when I have tested similarly for ```layout``` publish consisting of ```products``` with some missing attributes it failed. Not sure if ok or not in this case...


Screenshot 2025-01-29 134953

Interestingly it gives warning even if just some of the loaded asset in the scene is missing some attributes (and not been part of the publish ```layout``` instance) ...maybe not necessary to care about any products which are not directly in the publish instance??


Screenshot 2025-01-29 134810


Just some things to consider...if its unrelated then good for approval I guess.

BigRoy:
@LiborBatek please test again with the Extract Layout.

LiborBatek (approved):
Now all works even for ```Layout``` product consisting of loaded products missing some attribs...

LGTM
Screenshot 2025-02-20 114831


Screenshot 2025-02-20 114846

Issue #181, opened 2024-11-18
AY-7165_Flatten Layouts on publish as JSON


antirotor:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

When layout is published in JSON, if the layout contains other layouts, loading such JSON in Unreal fails because the loader cannot expand nested layout representations (they'll be as maya files).

### Suggested implementation?

Solution would be to have an option for publishing json, that it would flatten the layouts hierarchy. This is lossy operations, so it should be disabled by default. 

### Describe alternatives you've considered:

It could be also solved by refreshing **assembly** product.

PR #189, opened 2024-12-02
Extract Layout: Flatten Layouts on publish as JSON 

moonyuet:
## Changelog Description
This ticket is to resolve the json exported from Extract Layout cannot be correctly loaded in Unreal when the layout instance contains another loaded layout product.
Resolve https://github.com/ynput/ayon-maya/issues/181

n/a

## Testing notes:
1. Launch Maya
2. Extract Layout
3. Load Layout in Unreal/Maya
4. Should be all working.

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_layout.py
+++ b/client/ayon_maya/plugins/publish/extract_layout.py
@@ -6,7 +6,7 @@ from typing import List
 
 from ayon_api import get_representation_by_id
 from ayon_maya.api import plugin
-from ayon_maya.api.lib import get_highest_in_hierarchy
+from ayon_maya.api.lib import get_highest_in_hierarchy, get_container_members
 from maya import cmds
 from maya.api import OpenMaya as om
 
@@ -178,13 +178,13 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
         self.log.debug("Extracted instance '%s' to: %s",
                        instance.name, json_representation)
 
-    def get_containers_sub_assembly(self, containers, asset):
+    def get_containers_sub_assembly(self, containers, container_root):
         """Allow to collect the asset containers through sub-assembly workflow if
         there is a layout container
 
         Args:
             containers (list): Ayon asset containers
-            asset (str): asset name
+            container_root (str): the root of the asset container
 
         Returns:
             dict: container dict with the container and its related asset name
@@ -193,17 +193,15 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
         for existing_container in containers:
             if cmds.getAttr(f"{existing_container}.loader").lower() == "layoutloader":
                 for container in cmds.sets(existing_container, query=True):
-                    container_namespace = cmds.getAttr(f"{container}.namespace")
-                    container_rn = cmds.ls(f"{container_namespace}:*", references=True)
-                    if container_rn:
-                        associated_nodes = cmds.referenceQuery(container_rn, nodes=True)
-                    else:
-                        associated_nodes = cmds.ls(f"{container_namespace}:*")
-                    container_asset = get_highest_in_hierarchy(associated_nodes)[0]
-                    all_containers_set.update({container: container_asset})
+                    members = get_container_members(container)
+                    transforms = cmds.ls(members, transforms=True)
+                    roots = get_highest_in_hierarchy(transforms)
+                    root = next(iter(roots), None)
+                    if root is not None:
+                        all_containers_set[container] = root
             else:
-                container_asset = asset
-                all_containers_set.update({existing_container: container_asset})
+                all_containers_set[existing_container] = container_root
+
         return all_containers_set
 
     def create_transformation_matrix(self, local_matrix, local_rotation):
--- a/client/ayon_maya/plugins/publish/extract_layout.py
+++ b/client/ayon_maya/plugins/publish/extract_layout.py
@@ -87,7 +87,7 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
                 continue
 
             container_dict = self.get_containers_sub_assembly(containers, asset)
-            for container, asset in container_dict.items():
+            for container, container_root in container_dict.items():
                 representation_id = cmds.getAttr(
                     "{}.representation".format(container))
 
@@ -130,8 +130,10 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
                     "host": self.hosts
                 }
 
-                local_matrix = cmds.xform(asset, query=True, matrix=True)
-                local_rotation = cmds.xform(asset, query=True, rotation=True, euler=True)
+                local_matrix = cmds.xform(
+                    container_root, query=True, matrix=True)
+                local_rotation = cmds.xform(
+                    container_root, query=True, rotation=True, euler=True)
 
                 t_matrix = self.create_transformation_matrix(local_matrix, local_rotation)
 
--- a/client/ayon_maya/plugins/publish/extract_layout.py
+++ b/client/ayon_maya/plugins/publish/extract_layout.py
@@ -86,7 +86,7 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
                 self.log.warning("It may not be properly loaded after published") # noqa
                 continue
 
-            container_dict = self.get_containers_sub_assembly(containers, asset)
+            container_dict = self.process_container(containers)
             for container, container_root in container_dict.items():
                 representation_id = cmds.getAttr(
                     "{}.representation".format(container))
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
It does not trigger any error once loaded into ```UE``` or ```Maya``` when using nested ```layout``` product inside the ````layout```.

note: if using nested layout products and no other directly loaded assets/products within the layout used for loading....it results into empty UE map (but not triggering any error, which is in the scope of this PR I guess - not particulary an issue but worth mentioning and for consideration for future developement)

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +178,34 @@ def process(self, instance):
...
+        Args:
+            containers (list): Ayon asset containers
+            asset (str): asset name
+
+        Returns:
+            dict: container dict with the container and its related asset name
+        """
+        all_containers_set = {}
+        for existing_container in containers:
+            if cmds.getAttr(f"{existing_container}.loader").lower() == "layoutloader":
```
BigRoy, line 194:
```suggestion
            # Flatten layout loader to their individual loaded containers
            if cmds.getAttr(f"{existing_container}.loader").lower() == "layoutloader":
```

Also, why are we doing a `lower()`comparison? Don't we know the _exact_ plug-ins name?

```suggestion
            # Flatten container layout loader products to their individual loaded containers
            if cmds.getAttr(f"{existing_container}.loader") == "LayoutLoader":
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -84,80 +85,81 @@ def process(self, instance):
...
-                json_element["basis"].append(list(row))
-
-            json_element["rotation"] = {
-                "x": local_rotation[0],
-                "y": local_rotation[1],
-                "z": local_rotation[2]
-            }
-            json_data.append(json_element)
+            container_dict = self.get_containers_sub_assembly(containers, asset)
+            for container, asset in container_dict.items():
```
BigRoy, line 90: The `asset` variable here shadows the `asset` variable in the outer scope above, overriding it. Please provide a new unique variable name.

Also, everytime I read this code I'm very confused about what this _asset_ is - but all I can see is that it's intended to represent the "root" node transform of the container? I wonder if we should improve the variable somewhere or add some comments about this in the code.

moonyuet: `asset` variable mean the loaded asset which is hooked to the container data. maybe we can rename as `loaded_asset`

BigRoy: Still unsure what that means? What is the "loaded asset" - I have a feeling you're describing something in the scene but I have no knowledge myself of anything in Maya being known as 'an asset'. To me `asset` sounds very much like OpenPype "folder" and maybe it's actually referring to the folder? (which I don't think it does looking at the code).

To me it's more of a `container_root`, or `container_transform`. Also, note that _technically_ something loaded in _may_ have multiple root nodes, but I don't believe we're supporting that in the Layout Publishing and are always assuming loaded grouped containers where we use the 'group'  as the root node for that container.

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,32 @@ def process(self, instance):
...
+            containers (list): Ayon asset containers
+            container_root (str): the root of the asset container
+
+        Returns:
+            dict: container dict with the container and its related asset name
+        """
+        all_containers_set = {}
+        for existing_container in containers:
+            if cmds.getAttr(f"{existing_container}.loader").lower() == "layoutloader":
+                for container in cmds.sets(existing_container, query=True):
```
BigRoy, line 197: Technically not everything in the first container, may itself be a container (or even an object set for that matter).
With a loaded layout product 'most likely it is' but artists really break everything. So I'd at least filter to the data that we'd expect before doing anything with it.

Also, this initial set query - could also be replaced by `get_container_members`
```suggestion
                for container in get_container_members(existing_container):
                    # todo: confirm the object is actually a container?
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,32 @@ def process(self, instance):
...
+        Args:
+            containers (list): Ayon asset containers
+            container_root (str): the root of the asset container
+
+        Returns:
+            dict: container dict with the container and its related asset name
+        """
+        all_containers_set = {}
+        for existing_container in containers:
+            if cmds.getAttr(f"{existing_container}.loader").lower() == "layoutloader":
```
BigRoy, line 196: Could a layout loader - technically contain another layout loader, if so - this may need to just be recursive?

moonyuet: haven't thought of that before. Implement in [461e648](https://github.com/ynput/ayon-maya/pull/189/commits/461e64877655fe69afd7c2e2276bce8b975fd8e8)

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,33 @@ def process(self, instance):
...
+        Returns:
+            dict: container dict with the container and its related asset name
+        """
+        all_containers_set = {}
+        for container in containers:
+            if cmds.getAttr(f"{container}.loader") == "LayoutLoader":
+                # Flatten container layout loader products to their individual loaded containers
+                for member_container in get_container_members(container):
+                    # Recursively process each member container
+                    self.process_container(member_container)
```
BigRoy, line 199: This returns nothing - but the results should actually be 'updated' into the parent dict.

```suggestion
                    children_containers = self.process_container(member_container)
                    all_containers_set.update(children_containers)
```
I suppose something like this would be needed?

BigRoy: Also, `process_containers` takes a `list[str]` - not a single container, so this call is wrong to begin with?

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,34 @@ def process(self, instance):
         self.log.debug("Extracted instance '%s' to: %s",
                        instance.name, json_representation)
 
+    def process_container(self, containers):
+        """Allow to collect the asset containers through sub-assembly workflow if
+        there is a layout container
+
+        Args:
+            containers (list): Ayon asset containers
```
BigRoy, line 188:
```suggestion
            containers (list[str]): Ayon asset containers
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,34 @@ def process(self, instance):
         self.log.debug("Extracted instance '%s' to: %s",
                        instance.name, json_representation)
 
+    def process_container(self, containers):
```
BigRoy, line 183:
```suggestion
    def process_containers(self, containers):
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,34 @@ def process(self, instance):
...
+        Returns:
+            dict: container dict with the container and its related asset name
+        """
+        all_containers_set = {}
+        for container in containers:
+            if cmds.getAttr(f"{container}.loader") == "LayoutLoader":
+                # Flatten container layout loader products to their individual loaded containers
+                member_containers = get_container_members(container)
+                # Recursively process each member container
+                children_containers = self.process_container(member_containers)
```
BigRoy, line 199:
```suggestion
                children_containers = self.process_containers(member_containers)
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,34 @@ def process(self, instance):
...
 
+    def process_container(self, containers):
+        """Allow to collect the asset containers through sub-assembly workflow if
+        there is a layout container
+
+        Args:
+            containers (list): Ayon asset containers
+
+        Returns:
+            dict: container dict with the container and its related asset name
```
BigRoy, line 191:
```suggestion
            dict[str, str]: container mapping to its related root transform node
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,34 @@ def process(self, instance):
         self.log.debug("Extracted instance '%s' to: %s",
                        instance.name, json_representation)
 
+    def process_container(self, containers):
+        """Allow to collect the asset containers through sub-assembly workflow if
+        there is a layout container
```
BigRoy, line 185:
```suggestion
        there is a layout container.
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -84,80 +85,83 @@ def process(self, instance):
...
-            for row in b_matrix:
-                json_element["basis"].append(list(row))
-
-            json_element["rotation"] = {
-                "x": local_rotation[0],
-                "y": local_rotation[1],
-                "z": local_rotation[2]
-            }
-            json_data.append(json_element)
+            container_dict = self.process_container(containers)
```
BigRoy, line 89:
```suggestion
            container_dict = self.process_containers(containers)
```

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -176,6 +180,34 @@ def process(self, instance):
...
+                # Recursively process each member container
+                children_containers = self.process_container(member_containers)
+                all_containers_set.update(children_containers)
+            else:
+                members = get_container_members(container)
+                transforms = cmds.ls(members, transforms=True)
+                roots = get_highest_in_hierarchy(transforms)
+                root = next(iter(roots), None)
+                if root is not None:
+                    all_containers_set[container] = root
```
BigRoy, line 207: I wonder, is this more readable?
```suggestion
                if roots:
                    # Assume just a single root node
                    all_containers_set[container] = roots[0]
```

BigRoy (commented):
Looks good to me - just some docstring tweaks, and refactoring the method to `process_containers` since it processes multiple containers at once.

LiborBatek (commented):
Did produce some layout and when trying ```Load layout``` in the ```UE``` Im getting this error:

```
During load error happened on Product: "layoutMain" Representation: "json" Version: 1

Error message: 'NoneType' object has no attribute 'get_playback_start'

  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.0.10\ayon_core\tools\loader\models\actions.py", line 740, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.0.10\ayon_core\pipeline\load\utils.py", line 325, in load_with_repre_context
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\unreal_0.2.6\ayon_unreal\plugins\load\load_layout.py", line 305, in load
    generate_master_level_sequence(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\unreal_0.2.6\ayon_unreal\api\pipeline.py", line 1187, in generate_master_level_sequence
    e.get_asset().get_playback_start(),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_playback_start'
```

I have seen this already...back in time.

LiborBatek (approved):
I have re evaluated my findings with oficial UE addon version and all seems good now....

Issue #82, opened 2024-08-30
Maya USD: Publish animation with only point positions, transforms, bounding boxes as overrides


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Currently publishing USD with animation writes everything, like UVs - this bloats the output sizes but also introduces overrides you don't want from the animation products.

### How would you imagine the implementation of the feature?

We should strive to optimize, and allow writing e.g. only point positions, transforms, bounding boxes as **over** instead of **def**

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

PR #193, opened 2024-12-10
Implement Maya USD Export Chaser to Filter Properties

BigRoy:
## Changelog Description

Implement Maya USD Export Chaser to Filter Properties using a SideFX Houdini style pattern matching for all prim properties

For explanation on how the pattern works, see [SideFX Houdini Pattern Matching in Parameters](https://www.sidefx.com/docs/houdini/network/patterns.html) which it is loosely based on.

image

Results in:

image

```usda
#usda 1.0
(
    defaultPrim = "char_hero"
    metersPerUnit = 0.01
    upAxis = "Y"
)

def Xform "char_hero" (
    kind = "component"
)
{
    double3 xformOp:translate = (-0.4334395286605499, 0.8835878002001145, -0.6190157988997917)
    uniform token[] xformOpOrder = ["xformOp:translate"]

    def Mesh "pCube1"
    {
    }
}
```

Note how the mesh contains no properties at all - they are completely stripped from the written USD file.

### TODO

- [x] Improve attribute definition tooltip to describe the available pattern and maybe some examples.
- [x] Add to settings to enable/disable the plug-in and potentially set up a default value.

Added `ayon+settings://maya/publish/CollectMayaUsdFilterProperties`

image

## Testing notes:

1. A new attribute "USD Filter Properties" should appear on Maya USD instances
2. When any pattern is entered, then only those properties would be exported into the output file. For example:

- Only include xforms: `xformOp*`
- Everything but xforms: `* ^xformOp*`
- Everything but mesh point data: `* ^extent ^points ^faceVertexCounts ^faceVertexIndices ^primvars*`

### DIFF
--- a/client/ayon_maya/api/chasers/README.md
+++ b/client/ayon_maya/api/chasers/README.md
@@ -1,0 +1,14 @@
+# AYON Maya USD Chasers
+
+This folder contains AYON Maya USD python import and export chasers to be
+registered on Maya startup. These chasers have the ability to influence how
+USD data is imported and exported in Maya.
+
+For example, the Filter Properties export chaser allows to filter properties
+in the exported USD file to only those that match by the specified pattern
+using a SideFX Houdini style pattern matching.
+
+The chasers are registered in the `MayaHost.install` method on Maya launch.
+
+See also the [Maya USD Import Chaser documentation](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/commands/Readme.md#import-chasers) 
+and [Maya USD Export Chaser documentation](https://github.com/Autodesk/maya-usd/blob/dev/lib/mayaUsd/commands/Readme.md#export-chasers-advanced).
--- a/client/ayon_maya/api/chasers/export_filter_properties.py
+++ b/client/ayon_maya/api/chasers/export_filter_properties.py
@@ -1,0 +1,138 @@
+import re
+import fnmatch
+import logging
+from typing import List
+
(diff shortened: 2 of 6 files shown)


BigRoy:
This PR may also solve https://github.com/ynput/ayon-maya/issues/82 but only partially.
This PR does not:
- Strip prim definitions that are empty without properties
- Turn anything into `over` instead of `def`

LiborBatek:
yeah, its def very techy thing :) ...from artist pov at least (e.g. me :) 

I would like to recommend if that Ayon setting disabled (USD export chaser) then no new UI items shown in the publisher UI...if possible

and yes, disabled as defaults sounds right too!

moonyuet:
Checked in the published usd file: The xform data is there.
image
When the usd is imported back to Houdini, it appeared not to load anything(which is expected). When checking the geometry speadsheet, all xform attributes are loaded
image

moonyuet:
Tested with second scanerio where the xform is not included: (It shows all the vertex data of the meshes)
image
As the xform data is not exported, which means all the transforms of the meshes become zero when importing back to houdini
image
All the geometries would be at zero transform
image

This is how the usd export without the filter looks like, which means the filtering function works as expected.
image

moonyuet (approved):
See comments above

BigRoy:
> I would like to recommend if that Ayon setting disabled (USD export chaser) then no new UI items shown in the publisher UI...if possible
> 
> and yes, disabled as defaults sounds right too!

Correct - that should be the behavior now.

PR #198, opened 2024-12-13
Correct transform matrix data when extract layout 

moonyuet:
## Changelog Description
This PR is to make sure the transform matrix data are correct and not being zero when exporting layout.

n/a

## Testing notes:
1. Launch Maya
2. Extract Layout with/without layout containers


LiborBatek (approved):
Did test it and works correctly...

image

PR #199, opened 2024-12-16
Load Layout from the json exported from maya layout should have correct transform

moonyuet:
## Changelog Description
This PR is to fix the incorrect transformation being set across the asset when loading layout from the json exported by extract layout in maya. 
Resolve: https://github.com/ynput/ayon-maya/issues/200

n/a

## Testing notes:
1. Create Layout
2. Extract Layout
3. Load Layout
4. All transforms should be correct.

### DIFF
--- a/client/ayon_maya/plugins/load/load_layout.py
+++ b/client/ayon_maya/plugins/load/load_layout.py
@@ -145,10 +145,9 @@ class LayoutLoader(plugin.Loader):
             self._set_transformation(asset, transform)
         else:
             transform = element["transform_matrix"]
-            rotation = element["rotation"]
             # flatten matrix to a list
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
Did try multiple times and also with built package for maya addon but it doesnt work once loaded back into maya as JSON.

Assets having some ```Y axis``` offset as seen on the pic below (negative offset)

Screenshot 2024-12-16 121606

LiborBatek (approved):
Now all seems correct and with proper transforms. Been tested and compared to ```.ma``` layout repre. Also cherrypicked asset been compared and transforms matches 1:1

LGTM

image

Issue #102, opened 2024-09-11
AY-6502_dirmap not working


ynbot:
### Additional context:
()
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-6502](https://app.clickup.com/t/86bzz8cfd)_

PR #202, opened 2024-12-17
Remove `maya_` prefix from dirmap settings

BigRoy:
## Changelog Description

Remove `maya_` prefix from dirmap settings.

Should be tested together with https://github.com/ynput/ayon-core/pull/1061

Fix https://github.com/ynput/ayon-maya/issues/102

**Note:** Settings are not maintained with backwards compatibility! However, dirmapping didn't work anyway - so is that crucial? :)

## Testing notes:

1. Maya Dirmap should work

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -5,7 +5,7 @@ from ayon_server.settings import (
     ensure_unique_names,
 )
 from .imageio import ImageIOSettings, DEFAULT_IMAGEIO_SETTINGS
-from .maya_dirmap import MayaDirmapModel, DEFAULT_MAYA_DIRMAP_SETTINGS
+from .dirmap import DirmapModel, DEFAULT_DIRMAP_SETTINGS
 from .include_handles import IncludeHandlesModel, DEFAULT_INCLUDE_HANDLES
 from .explicit_plugins_loading import (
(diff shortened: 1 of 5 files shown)


server/settings/main.py
```diff
@@ -51,8 +51,8 @@ class MayaSettings(BaseSettingsModel):
     )
     ext_mapping: list[ExtMappingItemModel] = SettingsField(
         default_factory=list, title="Extension Mapping")
-    maya_dirmap: MayaDirmapModel = SettingsField(
-        default_factory=MayaDirmapModel, title="Maya dirmap Settings")
+    dirmap: DirmapModel = SettingsField(
```
iLLiCiTiT, line 54: Note, if anyone did use it, we probably should add settings overrides conversion (don't know if it did work tho?).

BigRoy: It didn't work - because it was using the wrong key in `ayon-core` all this time. Last time it would've worked was OpenPype then.

iLLiCiTiT: But it was used in `prepare_root_value` too, so people might have filled the values, or?

BigRoy: Should be implemented with https://github.com/ynput/ayon-maya/pull/202/commits/5b12acbcf12283afdfbbae505126614b8cb4e77d

iLLiCiTiT (approved):

PR #204, opened 2024-12-19
Store the render file directory as `stagingDir`

### LINKED ISSUE #194: ValidateRenderImageRule trigger failure after recent ayon-core staging dir update

## Current Behavior
`ValidateRendeerImageRule` now triggers as `stagingDir` is filled on the instance and it doesn't match expected value from `workspace.mel`.

Collected value is absolute path to temp dir, expected value is `renders/maya` (eg it expect rendered files to be stored relatively to workfile.

## Expected Behavior
Couple of approaches:
- Logic needs to be redone
- collection of `stagingDir` should be disabled for `render` product type in Maya
- disabling that validator

### Steps To Reproduce:

_No response_

### Version
`ayon-core` most likely from `1.0.11`

### Relevant log output:

```shell

```

BigRoy:
## Changelog Description
Store the render file directory as `stagingDir`

Fix #194

## Testing notes:

1. Test with PR https://github.com/ynput/ayon-core/pull/1066
2. With `ValidateRenderImageRule` plug-in enabled publishing renders should work fine.
3. Submitting renders should work as intended with both custom staging dir set and when disabled.

### DIFF
--- a/client/ayon_maya/plugins/publish/collect_render.py
+++ b/client/ayon_maya/plugins/publish/collect_render.py
@@ -257,6 +257,7 @@ class CollectMayaRender(plugin.MayaInstancePlugin):
             "source": filepath,
             "expectedFiles": full_exp_files,
             "publishRenderMetadataFolder": common_publish_meta_path,
+            "stagingDir": common_publish_meta_path,
             "renderProducts": layer_render_products,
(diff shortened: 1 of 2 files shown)


LiborBatek:
...also checked if ```custom staging dir``` was used and properly cleaned up....all good!

antirotor (approved):

PR #208, opened 2024-12-27
Clean-up and fix Alembic creator attributes

BigRoy:
## Changelog Description

- Remove `write_color_sets` and `write_face_sets` from creators where relevant.
    - Note: For model creator these are maintained, see the **todo** in code because there is special logic surrounding the model product type's creator toggle for "Write Face Sets" since https://github.com/ynput/ayon-maya/pull/37 by @moonyuet however it overlaps heavily with the "write face sets" toggle on the actual Extractor which is a **completely separate toggle** and currently for this model instance's face sets logic to work it is required for BOTH toggles to be enabled.
- Fix Create Animation and Create Pointcache defaults from settings for Include Parent Hierarchy and Include User Defined Attributes creator attributes. Previously the settings would not define the defaults at all.
- Add tooltip (and TODO) for `attr` and `attrPrefix` creator attributes for model product type - because those are essentially defined by the Alembic extractor already and hence currently **only** influence the USD export of the model product type (which in essence is **disabled** by default)

n/a

## Testing notes:

1. Writing color sets and writing face sets with model, pointcache and animation instances should still behave as intended.
2. Defaults in settings for Create animation and Create pointcache should be applied, in particular for "Include Parent Hierarchy" and "Include User Defined Attributes".
3. The Model product "attr" and "attrPrefix" should be somewhat clearer as to what it applies to. it should only apply to USD export, because Alembic export already exposes its own attribute definition for it.
4. The model product 'write face sets' should still behave as intended as per https://github.com/ynput/ayon-maya/pull/37

### DIFF
--- a/client/ayon_maya/plugins/create/create_animation_pointcache.py
+++ b/client/ayon_maya/plugins/create/create_animation_pointcache.py
@@ -8,7 +8,10 @@ from ayon_core.lib import (
 )
 
 
-def _get_animation_attr_defs(create_context):
+def _get_animation_attr_defs(
+        create_context,
+        include_user_defined_attributes,
+        include_parent_hierarchy=False):
     """Get Animation generic definitions."""
     defs = lib.collect_animation_defs(create_context=create_context)
     defs.extend(
@@ -22,7 +25,8 @@ def _get_animation_attr_defs(create_context):
                 tooltip=(
                     "Whether to include parent hierarchy of nodes in the "
                     "publish instance."
-                )
+                ),
+                default=include_parent_hierarchy
             ),
             BoolDef(
                 "includeUserDefinedAttributes",
@@ -30,7 +34,8 @@ def _get_animation_attr_defs(create_context):
                 tooltip=(
                     "Whether to include all custom maya attributes found "
                     "on nodes as attributes in the Alembic data."
-                )
+                ),
+                default=include_user_defined_attributes
             ),
         ]
     )
@@ -97,7 +102,9 @@ class CreateAnimation(plugin.MayaHiddenCreator):
         return node_data
 
     def get_instance_attr_defs(self):
-        return _get_animation_attr_defs(self.create_context)
+        return _get_animation_attr_defs(self.create_context,
+                                        self.include_user_defined_attributes,
+                                        self.include_parent_hierarchy)
 
 
 class CreatePointCache(plugin.MayaCreator):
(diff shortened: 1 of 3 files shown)


moonyuet:
I did some simple test and checked the code. Looks good and works as expected.
image
But I do have a question whether there is some use case for writing face sets when we export the assets in animation or point cache(Maybe this is for @LiborBatek). It seems that it is used only for exporting the model product from maya to unreal (or SP) for the current stage.

BigRoy:
> But I do have a question whether there is some use case for writing face sets when we export the assets in animation or point cache(Maybe this is for @LiborBatek). It seems that it is used only for exporting the model product from maya to unreal (or SP) for the current stage.

I agree - there may be merit in it; however, I think it should be a separate PR and I'm thinking it should move from Creator option to the Extractor option and become its own dedicated sub-toggle of `writeFaceSets` so that you can write face sets regularly without the enforced face shader assignments, and then can enable the "force shader assignment to per face assignment for writing face sets" separately.

I'm just not sure how well that is currently exposed to the model product type to make that really easy to add as optional toggle that way - but that way the logic could be shared between all product types, etc.

moonyuet:
> > But I do have a question whether there is some use case for writing face sets when we export the assets in animation or point cache(Maybe this is for @LiborBatek). It seems that it is used only for exporting the model product from maya to unreal (or SP) for the current stage.
> 
> I agree - there may be merit in it; however, I think it should be a separate PR and I'm thinking it should move from Creator option to the Extractor option and become its own dedicated sub-toggle of `writeFaceSets` so that you can write face sets regularly without the enforced face shader assignments, and then can enable the "force shader assignment to per face assignment for writing face sets" separately.
> 
> I'm just not sure how well that is currently exposed to the model product type to make that really easy to add as optional toggle that way - but that way the logic could be shared between all product types, etc.
> 
> Either way, it can be a separate PR - it's new functionality that doesn't exist currently regardless for the animation and pointcache product type.

Indeed. We can create a new issue for that if there is any need for that functionality.

moonyuet (approved):

LiborBatek (approved):
LGTM.

I have experienced once some cosmetic glitch on legacy ```model``` instance in the Publisher UI as seen below...however once recreated, all good. The default values for various properties been tested and propagation happening without issues too.
Screenshot 2025-01-03 134831

BigRoy:
> I have experienced once some cosmetic glitch on legacy `model` instance in the Publisher UI as seen below...however once recreated, all good. The default values for various properties been tested and propagation happening without issues too.

This isn't a visual glitch - but I believe an intentional feature by @iLLiCiTiT 

Creator attributes that do not exist anymore in the attribute definitions will still be listed in the UI for clarity - but only with their json value - not with any attribute definition.

Whether that's preferred to remain visible or not I'll leave up to you to debate ;)

PR #210, opened 2025-01-07
Load File: Do not set the <UDIM> token explicitly

BigRoy:
## Changelog Description

Load File: Do not set the `<UDIM>` token explicitly.

Fixes images to be listed unresolved by Maya:
image

Also fixes "asset dependencies" for Deadline submissions.

## Testing notes:

1. Load a published UDIM sequence (from substance painter)
2. List the file node in the Maya File Path Editor, it should appear green and not red.
3. Publish scene render to deadline with "asset dependencies" enabled - the file asset dependency should be computed correctly.


moonyuet (approved):
Tested in my side. Looks good. Code also looks good for me.
image

PR #217, opened 2025-01-13
AY-7320 Skip meshes without faces in explicit face set member assignments

BigRoy:
## Changelog Description

1. Skip meshes without faces in explicit face set member assignments
2. Consider only doing the face assignment conversion on meshes actually in the export, instead of all that happen to be in the `instance`

- Avoids a `No object matches name` error when trying to force face shader assignments for Write Face Sets if a mesh has no faces at all.
- Also fixes the forced face set assignments to process all meshes in the instance, now it only includes e.g. meshes that are inside the `out_SET` for animation instances.

## Testing notes:

1. Export a mesh without faces (yeah, that'd still be a problem!) with Write Face Sets with this conversion enabled.
2. It should not error on our code.

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4407,6 +4407,17 @@ def force_shader_assignments_to_faces(shapes):
             if "." not in member:
                 # Convert to face assignments
                 member = f"{member}.f[*]"
+                if not cmds.objExists(member):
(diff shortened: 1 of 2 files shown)


moonyuet (approved):
Code looks good for me. Did the test with published in maya and imported in unreal. The face sets for shader still kept in Unreal as expected.
image

PR #220, opened 2025-01-15
Load VDB to Redshift: Create and assign a RedshiftStandardVolume material on load

BigRoy:
## Changelog Description

Create and assign a RedshiftStandardVolume material on loading VDB to Redshift.

image
image

This also fixes an issue where on load Redshift would sometimes report a warning that the file does not exist (because Maya still had to initialize the `useFrameExtension` behavior). This PR now directly forces maya to evaluate that and avoids the warning like:
```
// Warning:  [Redshift] Failed to find volume file C:/projects/ayontest/asset/char_hero/publish/vdbcache/vdbcacheMain/v005/ynts_char_hero_vdbcacheMain_v005.0001.vdb. Volume will be empty
```

The option can be disabled on the Load action via the option box on the action itself.

This avoids the issue that when loading the VDB that by default it would NOT display in the render - usually Redshift would log that it had no shader and hence be skipped. Now, a shader is created and assigned and the field channel to use is guessed from the available grid names: use density if available, otherwise use first grid name which may save time as to debug "why doesn't it render?" and typing a valid channel name manually by the artist.

It does introduce potential follow up issue:

1. These materials are unmanaged (not in the container currently), and removing the Volume from the scene via scene inventory will keep the material around. If we were to add it to the container we get to the downside that the material may be used by the artist for other objects in the scene. The question is, which of the two behaviors would be worse?
2. Each load of a Redshift Volume would now generate a new shader in the scene. It may be unwanted.

---

The creation will also be logged:
```
// ProductLoader : Created RedshiftStandardVolume: 'rsStandardVolume12' using channel 'surface'
```

## Testing notes:

1. Load Volume to Redshift.
2. With the material applied it should directly be able to render.


LiborBatek (changes requested):
I have used tray publisher to ingest some VDB cache file from internet (Embergen)

and Im getting this error once using ```Ayon>Load VDB (Redshift)``` I guess its because it has just ```density``` grid inside not ```color```

```
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.0.13\ayon_core\tools\loader\models\actions.py", line 740, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.0.13\ayon_core\pipeline\load\utils.py", line 325, in load_with_repre_context
  File "c:\Work\REPO\ayon-maya\client\ayon_maya\plugins\load\load_vdb_to_redshift.py", line 94, in load
    self._create_default_redshift_volume_shader(volume_node)
  File "c:\Work\REPO\ayon-maya\client\ayon_maya\plugins\load\load_vdb_to_redshift.py", line 175, in _create_default_redshift_volume_shader
    cmds.connectAttr(f"{material}.outColor",
RuntimeError: The source attribute 'unknown1.outColor' cannot be found.```



once assigned the shader manually and inspect the asset (data is ok)


![Screenshot 2025-01-17 103416](https://github.com/user-attachments/assets/d5574597-8928-4916-9ffd-6b0b3e521cc6)

**Comment by BigRoy** [comment]
@LiborBatek what version of Redshift are you using. I'm assuming `RedshiftStandardVolume` does not exist yet in that version?

**Comment by LiborBatek** [comment]
> @LiborBatek what version of Redshift are you using. I'm assuming `RedshiftStandardVolume` does not exist yet in that version?

RS 3.5.17

**Review by LiborBatek — APPROVED** [review]
Worked for others...so LGTM

File: client/ayon_maya/plugins/load/load_vdb_to_redshift.py [hunk 1]
```diff
@@ -141,3 +153,41 @@ def _set_path(grid_node,
 
         # Set file path
         cmds.setAttr(grid_node + ".fileName", path, type="string")
+
+        # Force refresh with the use frame extension
+        # This makes sure we can directly retrieve the `.gridNames` attribute
+        # and avoids potential 'Failed to find volume file' warnings that
+        # appear once on load whe Maya has not yet initialized use frame
```
BigRoy, line 160:
```suggestion
        # appear once on load when Maya has not yet initialized use frame
```

PR #225, opened 2025-01-27
Improve validation reports for look validations

BigRoy:
## Changelog Description

Improve validation reports for look validations

See the commit messages for more details.

Some examples:

image
image
image

## Testing notes:

1. Create a Maya look and publish should work
2. When assigning a new material without saving (no cbId) the validation message should be nice.
3. When publishing look where a mesh has default material, it should be disallowed and easy for artist to solve.
4. When publishing a look with texture where no filepath is set (or files are missing) validation message should be clear.

### DIFF
--- a/client/ayon_maya/plugins/publish/help/validate_node_ids.xml
+++ b/client/ayon_maya/plugins/publish/help/validate_node_ids.xml
@@ -15,11 +15,6 @@ save your scene again.
 After that restart publishing with Reload button.
 </description>
 <detail>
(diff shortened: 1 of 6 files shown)


LiborBatek (approved):
I have followed all testing steps and its working nicely for all circumstances and inform user in a clearer way whats happening and why!

Good to go!

One example when empty file texture node used:

Screenshot 2025-01-28 161217

Nice and clear!

PR #227, opened 2025-01-29
Replace <br> in validator reports for new lines

BigRoy:
## Changelog Description

Replace <br> in validator reports for new lines log reports

Since https://github.com/ynput/ayon-core/pull/1104 the `<br>` would also display directly in the publisher UI report - which would appear cluttered. However, even without that PR the command line output would've also had the `<br>` in there so removing them is the correct fix.

image

image

image

## Testing notes:

1. Validate Mesh UV output should look good
2. Validate Transform Naming Suffix report should look good `ayon+settings://maya/publish/ValidateTransformNamingSuffix`
3. Validate Transform Zero report should look good. `ayon+settings://maya/publish/ValidateTransformZero`

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_mesh_has_uv.py
+++ b/client/ayon_maya/plugins/publish/validate_mesh_has_uv.py
@@ -75,13 +75,13 @@ class ValidateMeshHasUVs(plugin.MayaInstancePlugin,
         invalid = self.get_invalid(instance)
         if invalid:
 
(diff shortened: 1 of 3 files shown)


LiborBatek (approved):
Following the testing steps, all 1, 2 , 3 occasion worked well and formatting was accurate/ nice to look at.

Havent spotted any issues, so good to go!

One example from the testing here:

Screenshot 2025-01-30 101127

Issue #231, opened 2025-02-04
Extract Review fails if audio file used can't be found on disk (or has & character in filename)


BigRoy:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior

In Maya, when publishing a review it will get the audio track from the current scene's timeline [here](https://github.com/ynput/ayon-maya/blob/6505cb7c494fadbf5eb71233ff84ee43aa7a8abe/client/ayon_maya/plugins/publish/collect_review.py#L136-L170).

If that points to a file with `&` in the filename, then the FFMPEG command will fail:

```
[in#1 @ 00000247ac8f62c0] Error opening input: No such file or directory
Error opening input file path/to/filename.
Error opening input files: No such file or directory
'tail.wav' is not recognized as an internal or external command,
operable program or batch file.
```

It would execute the ffmpeg command with the audio waveform file linked like this:
```
ffmpeg.exe -apply_trc gamma22 -start_number 1001 -framerate 25.0 -to 15.0000000000 -i \\Temp\\pyblish_tmp_nze55ajj\\reviewMain.%d.png -to 15.0000000000 -vn -i /path/to/file&tail.wav -pix_fmt yuv420p -crf 18 -g 1 -y \\Temp\\pyblish_tmp_nze55ajj\\reviewMain_h264.mp4
```

It's the `&` symbol in the subprocess command for which FFMPEG starts to think a new command starts.

(This may need a fix in `ayon-core` however)

## Expected Behavior

Allow the subprocess command to succeed (this should just be allowed).
Or if impossible, we should put in a validator to avoid the issue to be hard to track down.

### Steps To Reproduce:

1. In Maya, load an audio wave file with `&` in filename.
2. Publish review.

### Version

Latest develop.

### What platform were you running when you found the bug?

Windows

### Relevant log output:

```shell

```

BigRoy:
I assume the same thing happens if the audio file pointed to by Maya doesn't actually exist on disk (but the error may be clearer then.)

BigRoy:
Reopening solely for the **non-existing file** remainder issue - but the worst offender for an existing file with `&` in filename is now fixed in core.

iLLiCiTiT:
> Reopening solely for the non-existing file remainder issue - but the worst offender for an existing file with & in filename is now fixed in core.

Isn't that issue of maya integration? Why should extract review handle missing file collected by host integration?

BigRoy:
[...]

PR #232, opened 2025-02-04
Ignore invalid audio filepaths with a warning instead of erroring during Extraction

BigRoy:
## Changelog Description

Allow to publish if for whatever reason the timeline is referencing an audio file that the machine can't find.

This would log a warning during collect, instead of failing during extraction.

**One may argue that instead of doing this that it _may_ make more sense to instead use a dedicated "validator" to validate the files exist and disallow it and warn the user.** A choice may need to be made about which makes more sense from a pipeline perspective.

Not sure what the preference is?

Currently it would allow publishing, but would show this logged message:
```
Timeline audio file not found on disk: '/path/to/nonexisting/file.wav'. Ignoring audio.
```
image

Fix https://github.com/ynput/ayon-maya/issues/231 for missing files.

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_maya/plugins/publish/collect_review.py
+++ b/client/ayon_maya/plugins/publish/collect_review.py
@@ -1,3 +1,5 @@
+import os
+
 import ayon_api
 import pyblish.api
 from ayon_core.pipeline import KnownPublishError
@@ -166,5 +168,18 @@ class CollectReview(plugin.MayaInstancePlugin):
 
                     if start_audio <= end_frame and end_audio > start_frame:
                         audio_data.append(get_audio_node_data(node))
+
+            # Exclude audio data with filenames that can't be found with a
+            # warning to avoid a harder error on e.g. Extract Review
+            valid_audio_data = []
+            for audio in audio_data:
+                if not os.path.exists(audio["filename"]):
+                    self.log.warning(
+                        "Timeline audio file not found on disk: '{}'. "
+                        "Ignoring audio.".format(audio["filename"])
+                    )
+                    continue
+                valid_audio_data.append(audio)
+            audio_data = valid_audio_data
 
             instance.data["audio"] = audio_data


LiborBatek:
I will give it a go... and test it

Speaking of the preference, I think we should have validator for it so it should fail when doing ```Review``` and maybe just warning if outputting products directly not affected? e.g. ```animation``` or ```pointcache``` ?

Just my 2 cents.

BigRoy:
> I will give it a go... and test it
> 
> Speaking of the preference, I think we should have validator for it so it should fail when doing `Review` and maybe just warning if outputting products directly not affected? e.g. `animation` or `pointcache` ? As there is still possibility to turn that validator as `Optional`...which might be more safe in the end (having that validation)
> 
> Just my 2 cents.

So this should only be related to "review" as far as I know and the audio being added to the playblast.
However, if the audio file wouldn't be found - then technically it also wouldn't play in Maya itself.
So doing the publish would basically still be the same as what the user is seeing themselves in Maya.
We'd just be invalidating then that the user has some dangling unused audio node inside the workfile.

The question then is: do we care to tell them?

E.g. it could be that someone happens to reference a scene, that contains an audio file, that has no filepath... and the user had never intended to _actually_ use the file node. The user can't easily remove it - are they now forced to put an audio track in just to review? (Unless they disable the validation). Or should we just warn them "hey, there's some dangling audio file node in the scene" and let them continue with a warning?

LiborBatek (approved):
Works as expected....when audio clip been missing just warning message been present and publish finnishes ok.
Screenshot 2025-02-06 113512

BigRoy:
> I will give it a go... and test it
> 
> Speaking of the preference, I think we should have validator for it so it should fail when doing `Review` and maybe just warning if outputting products directly not affected? e.g. `animation` or `pointcache` ? As there is still possibility to turn that validator as `Optional`...which might be more safe in the end (having that validation)
> 
> Just my 2 cents.

If a validator is needed - let's make a separate issue.

PR #241, opened 2025-02-19
Allow importing MayaUSDReferenceLoader + do not show the "Import Reference" option on non-supported containers

BigRoy:
## Changelog Description

Allow importing MayaUSDReferenceLoader + do not show the "Import Reference" option on non-supported containers

Load like:

image

Now also shows on Maya USD Reference:

image

Does not show on non-references:

image

Still shows for e.g. referenced alembics:

image

## Testing notes:
1. In Scene Inventory Maya USD Reference can now be "imported" as inventory action.
2. Non-reference loader containers now also should not show the "Import Reference" action.


LiborBatek (approved):
Works for me...

No issues spotted...

PR #242, opened 2025-02-19
Extract FBX: Export only once

BigRoy:
## Changelog Description

Extract FBX: Export only once

Seems like somehow we were exporting twice?

## Testing notes:

1. Exporting FBX `model` or `camera` (if enabled in settings) should still work as intended

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_fbx.py
+++ b/client/ayon_maya/plugins/publish/extract_fbx.py
@@ -43,8 +43,6 @@ class ExtractFBX(plugin.MayaExtractorPlugin):
         # Export
         with maintained_selection():
             fbx_exporter.export(members, path)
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
I cant tell if the publishing do the extraction just once but I can tell it correctly produced my new ```animation``` version and subjectively even faster?

So I guess good to go!

image

LiborBatek:
Haha good one! Someone updated the testing steps :) ...ok, I have done testing also for those guys ```model``` and ```camera``` both are working fine, and *cough cough faster too? :)

BigRoy:
@moonyuet feel free to just do a code review, and merge it when you can.

moonyuet (approved):

PR #248, opened 2025-03-07
Create Review: Use `get_attr_defs_for_instance` for attribute defaults based on settings

BigRoy:
## Changelog Description

Create Review: Use `get_attr_defs_for_instance` for attribute defaults based on settings

Also related to https://github.com/ynput/ayon-core/issues/1173 because it seems this method does not have easy access to `create_context` nor the logger from `Creator` itself.

When testing, beware of subtle edge case bug for which a separate fix exists here: https://github.com/ynput/ayon-maya/pull/247

## Testing notes:

1. Configure Extract Playblast defaults in settings
2. The attribute defaults in publisher UI should adhere to them for resolution width/height, isolate and image plane.

### DIFF
--- a/client/ayon_maya/plugins/create/create_review.py
+++ b/client/ayon_maya/plugins/create/create_review.py
@@ -1,7 +1,4 @@
-import json
-
-from maya import cmds
-import ayon_api
+import logging
 
 from ayon_maya.api import (
     lib,
@@ -12,7 +9,6 @@ from ayon_core.lib import (
     NumberDef,
     EnumDef
 )
-from ayon_core.pipeline import CreatedInstance
 
 TRANSPARENCIES = [
     "preset",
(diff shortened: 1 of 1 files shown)


client/ayon_maya/plugins/create/create_review.py
```diff
@@ -35,69 +31,18 @@ class CreateReview(plugin.MayaCreator):
...
-                "review_height": preset["Resolution"]["height"],
-                "isolate": preset["Generic"]["isolate_view"],
-                "imagePlane": preset["ViewportOptions"]["imagePlane"],
-                "panZoom": preset["Generic"]["pan_zoom"]
-            }
-            for key, value in mapping.items():
-                creator_attribute_defs_by_key[key].default = value
-
-            self._add_instance_to_context(instance)
+    @classmethod
```
iLLiCiTiT, line 34: Why is this classmethod?

BigRoy: My bad! I was clearly sleeping when drafting this PR.

LiborBatek (approved):
I have tested with several profiles each having distinct settings and all been correctly propagated when creating the ```review``` instance and also marked as override when not matching on existing publish instance...also when reset it correctly adhere to the ayon settings.

LGTM

PR #249, opened 2025-03-08
Extract Playblast: Task types settings show dropdown instead of text field

BigRoy:
## Changelog Description

Fix `task_types: list[str]` not appearing as multiselection dropdown

No idea why this fixes it - there may be some schema lookup clash on `ProfilesModel`?

It seems there may have been a clash in model type name with another object name, [likely this one](https://github.com/ynput/ayon-maya/blob/67abf92300497839ef1239f02d57df89b76a3eeb/server/settings/workfile_build_settings.py#L37)... and for whatever reason that made it so that the frontend stopped displaying the correct dropdown for it.

This is related to https://github.com/ynput/ayon-frontend/issues/1057

---

Before:

Image

After:

image

## Testing notes:
1. Upload addon
2. Setting `ayon+settings://maya/publish/ExtractPlayblast/profiles/0/task_types` should display as multiselection dropdown.

### DIFF
--- a/server/settings/publish_playblast.py
+++ b/server/settings/publish_playblast.py
@@ -252,7 +252,7 @@ class CapturePresetSetting(BaseSettingsModel):
         title="Camera Options")
 
 
-class ProfilesModel(BaseSettingsModel):
+class PlayblastProfilesModel(BaseSettingsModel):
     _layout = "expanded"
     task_types: list[str] = SettingsField(
         default_factory=list,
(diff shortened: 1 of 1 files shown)


BigRoy:
@martastain was so kind to explain. I asked "Am I being stupid?":

> Pydantic uses the class name in the json schema

As such, class names **should not conflict** within the addon.

LiborBatek (approved):
I can confirm that ```task type``` drop down been present and working,

tested with  distinct profiles via task type and different viewport 2.0 configuration and all working nicely.

Good to go!

PR #251, opened 2025-03-10
Fix writing redshift proxy filenames with number at end of filename (e.g. variant `env050`)

### LINKED ISSUE #114: AY-6616_Redshift Proxy Extract fails with variant name ending with digits

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1276455903692066916/1276456239781384284)

_This issue was automatically created from Clickup ticket [AY-6616](https://app.clickup.com/t/86c02uqr6)_

BigRoy:
## Changelog Description

Fix writing redshift proxy filenames with number at end of filename, e.g. instance name had `env050` as variant). Redshift would use the 050 as the frame number instead of appending the `.####` frame digits. We now write with `####` in the filepath which Redshift uses to write the frame number to as was intended.

Fix #114

## Testing notes:

1. Create redshift proxy instance with variant name `env050` (or anything that ends with digits)
2. Publishing should work.

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_redshift_proxy.py
+++ b/client/ayon_maya/plugins/publish/extract_redshift_proxy.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """Redshift Proxy extractor."""
+from __future__ import annotations
 import os
+from typing import Union
 
 from ayon_maya.api.lib import maintained_selection, renderlayer
 from ayon_maya.api import plugin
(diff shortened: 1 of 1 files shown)


moonyuet (changes requested):
When the animation off, the integrate error out due to the frame token isn't replaced by the frame range(current frame).
But animation on works perfectly
See the report for more detail [publish-report-250311-16-16.json](https://github.com/user-attachments/files/19178688/publish-report-250311-16-16.json)

moonyuet (approved):
Looks good now.
image
image

PR #253, opened 2025-03-11
Render Settings: Refactor setting of Redshift GI engine - do not ever set value of `0`

BigRoy:
## Changelog Description

Refactor setting of Redshift GI engine - do not ever set value of `0`

(which would've previously happened if any of the entries was set to "None" in settings). Currently "None" would set it to the Redshift default values where it is:

- None for Primary Engine is: Brute Force
- None for Secondary GI Engine is: Irradiance Point Cloud

Which leaves the question: Why do we still need the `None` field? Or should the `None` field be applied as "Do not set it at all?"

## Testing notes:

1. Settings Primary GI Engine "None" and Secondary GI Engine "Irradiance Point Cloud" should not give unexpected results.
2. There was a report that saving the scene with the render settings set like that and reopening it, that Redshift would display the settings differently with the "Irradiance Cache" section under the GI settings in Redshift settings appearing greyed out (as if it's currently set to an invalid value)

### DIFF
--- a/client/ayon_maya/api/lib_rendersettings.py
+++ b/client/ayon_maya/api/lib_rendersettings.py
@@ -197,22 +197,22 @@ class RenderSettings(object):
         # update the AOV list
         mel.eval("redshiftUpdateActiveAovList")
 
-        rs_p_engine = redshift_render_presets["primary_gi_engine"]
-        rs_s_engine = redshift_render_presets["secondary_gi_engine"]
+        # Irradiance Point cloud: 2 (only available for secondary GI engine)
+        # Irradiance Cache: 3 (only available for primary GI engine)
+        # Brute Force: 4
+        rs_p_engine = int(redshift_render_presets["primary_gi_engine"])
+        rs_s_engine = int(redshift_render_presets["secondary_gi_engine"])
+        if rs_p_engine == 0:
+            # reset the primary GI Engine as default: Brute Force
+            rs_p_engine = 4
+        if rs_s_engine == 0:
+            # reset the secondary GI Engine as default: Irradiance Point cloud
+            rs_s_engine = 2
 
-        if int(rs_p_engine) or int(rs_s_engine) != 0:
-            cmds.setAttr("redshiftOptions.GIEnabled", 1)
-            if int(rs_p_engine) == 0:
-                # reset the primary GI Engine as default
-                cmds.setAttr("redshiftOptions.primaryGIEngine", 4)
-            if int(rs_s_engine) == 0:
-                # reset the secondary GI Engine as default
-                cmds.setAttr("redshiftOptions.secondaryGIEngine", 2)
-        else:
-            cmds.setAttr("redshiftOptions.GIEnabled", 0)
-
-        cmds.setAttr("redshiftOptions.primaryGIEngine", int(rs_p_engine))
-        cmds.setAttr("redshiftOptions.secondaryGIEngine", int(rs_s_engine))
+        gi_enabled: bool = bool(rs_p_engine or rs_s_engine)
+        cmds.setAttr("redshiftOptions.GIEnabled", gi_enabled)
+        cmds.setAttr("redshiftOptions.primaryGIEngine", rs_p_engine)
+        cmds.setAttr("redshiftOptions.secondaryGIEngine", rs_s_engine)
 
         additional_options = redshift_render_presets["additional_options"]
         ext = redshift_render_presets["image_format"]


client/ayon_maya/api/lib_rendersettings.py
```diff
@@ -197,22 +197,22 @@ def _set_redshift_settings(self):
...
-        cmds.setAttr("redshiftOptions.primaryGIEngine", int(rs_p_engine))
-        cmds.setAttr("redshiftOptions.secondaryGIEngine", int(rs_s_engine))
+        # Irradiance Point cloud: 2 (only available for secondary GI engine)
+        # Irradiance Cache: 3 (only available for primary GI engine)
+        # Brute Force: 4
+        rs_p_engine = int(redshift_render_presets["primary_gi_engine"])
+        rs_s_engine = int(redshift_render_presets["secondary_gi_engine"])
+        if rs_p_engine == 0:
+            # reset the primary GI Engine as default: Brute Force
+            rs_p_engine = 4
```
moonyuet, line 207: I am thinking if it would be better to use the current setting for primary/secondary GI Engines to allow some freedom for users to use their own preferences(just in case they want to test something with that).
```suggestion
            rs_p_engine = int(cmds.getAttr("redshiftOptions.primaryGIEngine"))
```

client/ayon_maya/api/lib_rendersettings.py
```diff
@@ -197,22 +197,22 @@ def _set_redshift_settings(self):
...
+        if rs_s_engine == 0:
+            # reset the secondary GI Engine as default: Irradiance Point cloud
+            rs_s_engine = 2
```
moonyuet, line 210:
```suggestion
            rs_s_engine = int(cmds.setAttr("redshiftOptions.secondaryGIEngine"))
```

moonyuet:
I have tested the current code in my side, I have commented on several things. Once it is answered and it can be re-tested and merged.

client/ayon_maya/api/lib_rendersettings.py
moonyuet: >I'd have expected myself having a separate "GI enabled" checkbox in settings makes more sense - so it's explicit about what is happening.

I agree with this as it would ease the confusion, if the purpose of `None` is not for users to set it by their own.

client/ayon_maya/api/lib_rendersettings.py
```diff
@@ -197,22 +197,14 @@ def _set_redshift_settings(self):
...
-                cmds.setAttr("redshiftOptions.primaryGIEngine", 4)
-            if int(rs_s_engine) == 0:
-                # reset the secondary GI Engine as default
-                cmds.setAttr("redshiftOptions.secondaryGIEngine", 2)
-        else:
-            cmds.setAttr("redshiftOptions.GIEnabled", 0)
-
-        cmds.setAttr("redshiftOptions.primaryGIEngine", int(rs_p_engine))
-        cmds.setAttr("redshiftOptions.secondaryGIEngine", int(rs_s_engine))
+        gi_enabled: bool = redshift_render_presets["gi_enabled"]
```
moonyuet, line 200:
```suggestion
        gi_enabled: bool = redshift_render_presets.get("gi_enabled", False)
```

moonyuet (changes requested):
It cannot find the key element `gi_enabled` from the render setting dict
image
```
Creator: Render (io.openpype.creators.maya.renderlayer)

Error: 'gi_enabled'

  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.1.4+dev\ayon_core\pipeline\create\context.py", line 2165, in _create_with_unified_error
    result = creator.create(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\ayon-addon_template\ayon-maya\client\ayon_maya\plugins\create\create_render.py", line 55, in create
    lib_rendersettings.RenderSettings().set_default_renderer_settings()
  File "D:\ayon-addon_template\ayon-maya\client\ayon_maya\api\lib_rendersettings.py", line 113, in set_default_renderer_settings
    self._set_redshift_settings()
  File "D:\ayon-addon_template\ayon-maya\client\ayon_maya\api\lib_rendersettings.py", line 200, in _set_redshift_settings
    gi_enabled: bool = redshift_render_presets["gi_enabled"]
                       ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'gi_enabled'
```

client/ayon_maya/api/lib_rendersettings.py
BigRoy: I am always so confused by this. The setting should be there because it's in the settings model.

Did you test without uploading the addon by any chance?

moonyuet: > I am always so confused by this. The setting should be there because it's in the settings model.
> 
> Did you test without uploading the addon by any chance?

I tried to clean up the maya addon and re-install again. Working now. It could be because I am using development mode.

moonyuet (approved):
Looks good now. 
All global illumination engine set as expected and if both engines are set to None in ayon setting, it would be turned off. 
image

client/ayon_maya/api/lib_rendersettings.py
BigRoy: @martastain @iLLiCiTiT do you know by any chance?

Say I deploy a new addon to the server with changed settings, would that new project settings directly get returned from the server when e.g. launching a new app. Or is there any reason to still fallback as if the setting would not exist?

If it were to update automatically, only thing I can think of why this may have failed for Kayla was running dev mode without having the revised addon uploaded to her server. (Do note that because in dev mode, it may of course be a  re-upload of an existing dev version).

In short, should this be `.get()` or can we rely on the keys existence through the Pydantic settings models?

BigRoy:
> Looks good now. All global illumination engine set as expected and if both engines are set to None in ayon setting, it would be turned off. image

@moonyuet this screenshot is **wrong** - these are the old settings. You didn't seem to have recreated the addon package and uploaded it to the server (and set it active in your dev bundle?)

It should look like:
image

client/ayon_maya/api/lib_rendersettings.py
iLLiCiTiT: It can happen only in dev mode when you don't upload the addon to server. It won't happen in release...

BigRoy: In that case, code should be fine as is. Thanks for confirming.

moonyuet (approved):
Looks good 
image
image
image

PR #259, opened 2025-03-24
Create Render: Use entity getters from `create_context`

BigRoy:
## Changelog Description

Create Render: Use entity getters from `create_context`

Collecting new render setup layers as new render instances should still work.

## Testing notes:

1. Collecting new render setup layers as new render instances should still work.


iLLiCiTiT (approved):

LiborBatek (approved):
I haven't experienced any issues while re-creating the ```render``` instance or just added/removed new renderlayer to the maya scene...all was reflected accordingly.

LGTM

image

PR #265, opened 2025-03-31
Load Matchmove: Fix load mel script with backslashes paths

BigRoy:
## Changelog Description

Load Matchmove: Fix load mel script with backslashes paths

## Additional review information

Fixes potential issue with backslashes in paths.

Related to [(private) internal conversation](https://discord.com/channels/517362899170230292/1356203288659230831).

## Testing notes:

1. Load matchmove script (.mel) should work

PR #268, opened 2025-04-04
Fix paths passed to MEL matchmove script loader

antirotor:
## Changelog Description
Wrong slashes caused `mel.eval` to fail to load and execute published MatchMove MEL file.

## Testing notes:
Publish Matchmove as MEL and load it it Maya

### DIFF
--- a/client/ayon_maya/plugins/load/load_matchmove.py
+++ b/client/ayon_maya/plugins/load/load_matchmove.py
@@ -1,4 +1,6 @@
 import runpy
+from pathlib import Path
+from typing import Optional
 
 from ayon_maya.api import plugin
 from maya import mel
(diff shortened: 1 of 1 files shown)


BigRoy:
I suppose this acts as a replacement to this: https://github.com/ynput/ayon-maya/pull/265

Anyway, client reported that this would still have issues on load. We should instead use `cmds.file(path, import=True, type="mel")` because for whatever reason that does not error on the `expression` command in the mel script generated from 3d equalizer directly after they enable `frameExtension` attribute on the image plane file node.

See [private conversation.](https://discord.com/channels/517362899170230292/1356203288659230831/1356960268638490694)

moonyuet:
I tested with loading matchmove product, it works for me without errors. We need to consider `cmds.file` approach instead of `mel.eval` to ensure no frame range issue found.

antirotor:
Right you are, I changed it to `cmds.file` and it works now as expected.

PR #270, opened 2025-04-25
Load Redshift Proxy: Support Alembic (abc) and USD files.

BigRoy:
## Changelog Description

Add support for Redshift Proxy to load `usd` and `abc` files as support has been added in recent Redshift releases 2025.4.0+

The file type toggle on `RedshiftProxyMesh` appears at the top of the attribute editor:
image

With this change, the attribute will set as needed so that USD and Alembic files read as intended.

## Testing notes:

1. Loading of RedshiftProxy should work in Redshift 2025.4.0+
2. Updating should also work.
3. In older releases the load option will appear but a warning will be logged in the script editor instead that it's not supported due to the redshift version.

### DIFF
--- a/client/ayon_maya/plugins/load/load_redshift_proxy.py
+++ b/client/ayon_maya/plugins/load/load_redshift_proxy.py
@@ -15,8 +15,9 @@ from ayon_maya.api.plugin import get_load_color_for_product_type
 class RedshiftProxyLoader(plugin.Loader):
     """Load Redshift proxy"""
 
-    product_types = {"redshiftproxy"}
-    representations = {"rs"}
+    product_types = {"*"}
+    representations = {"*"}
+    extensions = {"rs", "usd", "usda", "usdc", "abc"}
 
     label = "Import Redshift Proxy"
     order = -10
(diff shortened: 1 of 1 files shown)


client/ayon_maya/plugins/load/load_redshift_proxy.py
```diff
@@ -148,3 +153,40 @@ def create_rs_proxy(self, name, path):
             cmds.setAttr("{}.useFrameExtension".format(rs_mesh), 1)
 
         return nodes, group_node
+
+    def _set_rs_proxy_file_type(self, proxy: str, path: str):
+        """Set Redshift Proxy file type attribute based on input file."""
+
+        extension = os.path.splitext(path)[1].lower()
```
moonyuet, line 160: Would it be better to do this instead?
```suggestion
        extension = os.path.splitext(path)[-1].lower()
```

moonyuet (approved):
Tested it with the latest version of RS in Maya, looks good.
image

client/ayon_maya/plugins/load/load_redshift_proxy.py
BigRoy: Splitext always returns a two-tuple, so should be fine either way.

PR #274, opened 2025-05-06
updated anim export and imp

pavithiraraj:
## Changelog Description
This exports the anim data of the characters so when the next department picks up the file they will have the option to load the anim and continue animation

Have added collect_anim_curve so it exports the anim data and adds to the representation. And to load back the anim you can right click on the anim representation in the loader and click "Load Anim", if the reference is present it will add the anim data else it will reference the asset and add the anim data.
the anim data that is exported is in a json format, as the atom export was not able to load the anim if the ctrls are selected in a different order.

## Testing notes:
1. publish an layout or ra step in maya and load it using the loader.


MustafaJafar:
Could you rename this file `collect_anim_curve.py` to `export_anim_curve.py`

MustafaJafar:
Does anim export re-use an existing creator?

pavithiraraj:
> Could you rename this file `collect_anim_curve.py` to `export_anim_curve.py`

Have renamed the files.

pavithiraraj:
> Does anim export re-use an existing creator?

i dont get it, which existing creator?

client/ayon_maya/plugins/publish/export_anim_curve.py
```diff
@@ -0,0 +1,142 @@
...
+    families = ["animation"]
+    hosts = ["maya"]
+
+    def process(self, instance):
+        instance_data = instance.data
+        # Define output path
+        staging_dir = self.staging_dir(instance)
+        filename = "{0}.anim".format(instance_data['variant'])
+        out_path = os.path.join(staging_dir, filename)
+        controls = [x for x in instance_data['setMembers'] if x.endswith(":rigMain_controls_SET")]
```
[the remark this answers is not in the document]
moonyuet: the loaded rig product type can be with different variants, e.g. rigTest, rigTech_carA etc.

client/ayon_maya/plugins/load/load_anim.py
```diff
@@ -0,0 +1,139 @@
...
+    representations = {"anim"}
+
+    label = "Load Anim"
+    order = -5
+    icon = "code-fork"
+    color = "orange"
+
+    def load(self, context, name, namespace, data):
+        project_name = context['project']['name']
+        anim_file = context['representation']['attrib']['path']
```
moonyuet, line 28:
```suggestion
        anim_file = self.filepath_from_context(context)
```

client/ayon_maya/plugins/load/load_anim.py
```diff
@@ -0,0 +1,139 @@
...
+            for rep in representations:
+                if rep['name'] == 'ma':
+                    rep_id = rep['id']
+                    break
+            context = get_representation_context(project_name, rep_id)
+            data['attach_to_root'] = True
+            _plugin = ReferenceLoader()
+            _plugin.load(context=context, name=context['product']['name'],
+                                      namespace=name_space, options=data)
+        ctrl_set = pm.ls(name_space + ":rigMain_controls_SET")
```
moonyuet, line 57:
```suggestion
        ctrl_set = pm.ls(name_space + ":rigMain_controls_SET")
```
Keep in mind there is the loaded instance with different variant, e.g. rigTest_controls_SET, rigTechCarA_controls_SET

client/ayon_maya/plugins/load/load_anim.py
```diff
@@ -0,0 +1,139 @@
...
+from ayon_core.pipeline.load.utils import get_representation_context
+from ayon_maya.plugins.load.load_reference import ReferenceLoader
+import logging
+import pymel.core as pm
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+
+class AnimLoader(plugin.Loader):
```
moonyuet, line 15: You can inherit the loader from `plugin.ReferenceLoader` instead. 
Then you dont need to implement things in line 54-56

client/ayon_maya/plugins/load/load_anim.py
```diff
@@ -0,0 +1,139 @@
...
+        if not ctrl_set:
+            self.log.warning("No control set found in instance data")
+            return
+        ctrls = pm.listConnections(ctrl_set[0], source=1, type='transform')
+        if not ctrls:
+            self.log.warning("No controls found in instance data")
+            return
+        self.log.debug(f"ctrls: {ctrls}")
+        self.log.debug(f"namespace: {namespace}")
+        self.log.debug(f"anim_file: {anim_file}")
```
moonyuet, line 67:
```suggestion

```

client/ayon_maya/plugins/load/load_anim.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def load(self, context, name, namespace, data):
+        project_name = context['project']['name']
+        anim_file = context['representation']['attrib']['path']
+        self.log.info(f"anim_file: {anim_file}")
+        name_space = context['representation']['data']['context']['product']['name'].replace(
+            context['representation']['data']['context']['product']['type'], '')
+        if not cmds.namespace(exists=name_space):
+            assets = context['version']['data'].get("assets", [])
+            self.log.info(f"assets: {assets}")
```
moonyuet, line 34: You can take reference from other loaders in `/load` folder to get the namespace
```suggestion
        folder_name = context["folder"]["name"]
        namespace = namespace or lib.unique_namespace(
            folder_name + "_",
            prefix="_" if folder_name[0].isdigit() else "",
            suffix="_",
        )

```

client/ayon_maya/plugins/load/load_anim.py
```diff
@@ -0,0 +1,139 @@
...
+            current_asset = [asset for asset in assets if asset['asset_name'] in anim_file]
+            if not current_asset:
+                self.log.warning(f"Asset not found in version data.")
+                return
+            current_asset = current_asset[0]
+            asset_data = ayon_api.get_folder_by_name(project_name=project_name, folder_name=current_asset['asset_name'])
+            versions = ayon_api.get_last_version_by_product_name(project_name, "rigMain", asset_data['id'])
```
moonyuet, line 41:
```suggestion
            versions = ayon_api.get_last_version_by_product_name(project_name, "rigMain", asset_data['id'])
```
We can have the product name with different variants, e.g. rigTest, rigTechCarA etc.

client/ayon_maya/plugins/publish/export_anim_curve.py
```diff
@@ -0,0 +1,142 @@
...
+        reference_node = [x for x in pm.listReferences() if x.namespace == name_space][0]
+        self.log.info(f"controls: {controls}")
+        self.write_anim(objects=ctrls, filepath=os.path.realpath(out_path))
+        if "representations" not in instance.data:
+            instance.data["representations"] = []
+        representation = {
+            'name': 'anim', 'ext': 'anim',
+            'files': os.path.basename(out_path),
+            'stagingDir': staging_dir.replace("\\", "/")
+        }
```
moonyuet, line 40:
```suggestion
        representation = {
            'name': 'anim', 
            'ext': 'anim',
            'files': os.path.basename(out_path),
            'stagingDir': staging_dir.replace("\\", "/")
        }
```

pavithiraraj: This was starting to give errors so i removed this block to add representation, but still the anim got added to the published version, any idea as to why this is happening

pavithiraraj: i was getting that the representation already exists, so i tried renaming to anim_curve and that seems to create two entries
image

so once i removed it , it was back to normal

pavithiraraj: i am guessing that the publish code automatically adds the representation based on the instance, without the need to define it in the code.

client/ayon_maya/plugins/load/load_anim.py
```diff
@@ -0,0 +1,139 @@
...
+import maya.cmds as cmds
+import ayon_api
+from ayon_maya.api import plugin
+from ayon_core.pipeline.load.utils import get_representation_context
+from ayon_maya.plugins.load.load_reference import ReferenceLoader
+import logging
+import pymel.core as pm
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
```
BigRoy, line 12: No need to create a logger here - just pass the logger from the plug-in if you need one.

client/ayon_maya/plugins/publish/export_anim_curve.py
moonyuet: > i am guessing that the publish code automatically adds the representation based on the instance, without the need to define it in the code.

The Extractor plugins you have inherited from ` MayaExtractorPlugin` are also used for inheritance in other extractors(e.g. extract_obj.py etc). 
They are all with the similar implementation of the representation as showed below:
```python
if "representations" not in instance.data:
    instance.data["representations"] = []

representation = {
    'name': self.scene_type, # "anim" in your case
    'ext': self.scene_type, # "anim" in your case
    'files': filename, # os.path.basename(out__path) in your case
    "stagingDir": stagingdir,
}
instance.data["representations"].append(representation)

self.log.debug("Extracted instance '%s' to: %s" % (instance.name,
                                                    path))

```
It would be great if you can enclose the publish report by exporting report in Publisher UI and we can take the deep look to see where the errors come from. 
image

client/ayon_maya/plugins/publish/export_anim_curve.py
```diff
@@ -0,0 +1,142 @@
...
+        if not assets:
+            version_data["assets"] = []
+        asset_data = {
+            "namespace": name_space,
+            "path": reference_node.path.__str__(),
+            "asset_name": name_space.split("_rigMain_")[0],
+        }
+        version_data["assets"].append(asset_data)
+        instance_data["data"] = [name_space + ':' + reference_node.path.__str__()]
+        instance.data["assets"] = [name_space + ':' + reference_node.path.__str__()]
```
BigRoy, line 52: Seems like arbitrary data?
Why not store the rig's representation id instead? So that on load, you're also loading the same matching rig production version (and representation). It'd greatly simplify loading too, because you can just directly load it from the representation id. :)

BigRoy (changes requested):
So, I don't think this will end up being merged if it remains close to what it is now. However, the feature itself is great - so let's work towards making it mergable.

The issues;
- **Pymel dependendency.** The code is now allowed to use `pymel`. so please avoid using it.
- **Custom animation data file format.** This introduces a completely custom animation data file format - which I think is a bad thing. There are tools and formats out there that I'd rather rely on or mimic, like `atom` or matching e.g. `studiolibrary` export/import or even `animbot`'s anim export/import. Preferably it's a freely available one, and bonus points if it's by default included with Maya... but rolling our own I think is just plain overkill, plus not good for any interchangeable efforts down the line.

client/ayon_maya/plugins/publish/export_anim_curve.py
pavithiraraj: I have attached it already
[publish-report-250604-16-31.json](https://github.com/user-attachments/files/20590535/publish-report-250604-16-31.json)
But i found the issue, we had a separate package to do the anim export for now, and seems that addon was added to the bundle. (feeling dumb)

client/ayon_maya/plugins/publish/export_anim_curve.py
pavithiraraj: That sound like a solid plan, how do i get the representation id from the instance?

moonyuet: > That sound like a solid plan, how do i get the representation id from the instance? is it the inputRepresentations in the instance.data ?

You can take reference from https://github.com/ynput/ayon-maya/blob/9b026e716f5ecab3d7a7c9199eaaf2bc1364f947/client/ayon_maya/plugins/publish/extract_layout.py#L97-L115

pavithiraraj:
I have made most of the changes Can you have a look at it now, do you want me to change the way we write out the data, from a json to atom?
and do i need to remove pymel in the code?

pavithiraraj:
Ah ok , i will try and update this based on the layout loader, just curious, when they publish a next version, will the container know about the update? as we are loading the asset as a container and the anim data on top of that right?

BigRoy:
Closing this as communication has stalled and it's still pending quite some comments. Feel free to continue the efforts and re-open.

PR #281, opened 2025-05-27
Validate Maya Units: Support optional toggle in settings

BigRoy:
## Changelog Description

Support `optional` state in settings for Validate Maya Units: `ayon+settings://maya/publish/ValidateMayaUnits/optional`

Setting are automatically applied because it inherits from `MayaContextPlugin` which sets the `settings_category`
I suspect this may also _fix_ the "Enabled" state for the plug-in which likely previously didn't actually work to disable it.

## Testing notes:

1. All overrides for `ayon+settings://maya/publish/ValidateMayaUnits` should still work
2. Optional toggle should now also appear on "Context" entry in Publisher UI when setting is enabled.

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_maya_units.py
+++ b/client/ayon_maya/plugins/publish/validate_maya_units.py
@@ -35,21 +35,6 @@ class ValidateMayaUnits(plugin.MayaContextPlugin,
     )
     optional = False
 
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
Code looks good to me and tested successfully in Maya
image

PR #282, opened 2025-06-02
Bug: skip `None` values when imprinting

antirotor:
## Changelog Description
When imprinting, skip values that are `None` as there is no real way to convert them to Maya-native values, nor need for it. 

When Maya enounters `None` value when imprinting, it crashes with `TypeError: Unsupported type: <class 'NoneType'>`. I think that we can safely assume that if it is `None`, we don't need to store it.

## Testing notes:
- add some `None` value to instance data in a creator
- create instance in Maya using this creator


antirotor:
> How do you reproduce this in a production scenario?

It is actually very simple. Just add any custom attribute on the creator that is `None`:

```python
class TestCreator(plugin.MayaCreator)
    identifier = "io.ayon.creators.maya.test"
    label = "Test"
    product_type = "test"
    icon = "cube"
    default_variants = []

    some_great_attribute = None```

```

in this case `some_great_attribute` will trigger it. Admittedly, we shoud change what is actually imprinted, but that is IMO too deep now to handle. I had to make this change just to continue my work on https://github.com/ynput/ayon-core/issues/1297

I would argue that emitting any warning is now very much useless as this is usually lost in the console output and is of no use for the artist. Developer will check the code and see the comment. Also, it doesn't break any existing functionality because it would be crashing already with `None` values. Last point - I think handling `None` values should be done in the code, not on the stored data itself.

moonyuet:
Are we gonna merge this PR or?

BigRoy:
> Are we gonna merge this PR or?

I don't think we should, what Ondrej wrote:

> When Maya enounters `None` value when imprinting, it crashes with `TypeError: Unsupported type: <class 'NoneType'>`. I think that we can safely assume that if it is `None`, we don't need to store it.

I don't believe that to be true. It'd just be hiding an actual issue. If I were to imprints tens of attributes, and suddenly two were missing and I wouldn't know why it'd be odd. We can't just go and assume that if someone passed None that we can skip creating an attribute for it.

It should possibly error with a dedicated error - so some outside caller can do `try..except` on it.

Personally I'd want to hear more about the cases where the error occurs - why is `None` being passed?

@antirotor @YoussefMohamed-joe do you remember your use cases that you were trying to 'fix'?

BigRoy:
So I hit this issue today - and the cause was creating an instance that has no task set, which would set `task` to `None`.

Easy to reproduce, not a big problem - but definitely something worth fixing. In the case of the lacking `task` just not writing the attribute seems to work fine.

BigRoy:
@antirotor I've tweaked the logic here with https://github.com/ynput/ayon-maya/pull/282/commits/b012a1548ffdd1d1925b13a8d17074a780842c00

Can you let me know if ok for you?

BigRoy (approved):

BigRoy:
> Filtering can delete existing Maya attributes before skipping None values.

This is intentional, I do want them to get cleared out.

YoussefMohamed-joe:
Sorry for late reply, Yeah the same case happened to me but i don't really remember when did it happen as there were few use cases that made the same problem i remember handling it by making it accept none normally as an empty string and has been working well since. 
```python
elif value is None:  
    add_type = {"dataType": "string"}
    set_type = {"type": "string"}
    value = ""
```

Issue #90, opened 2024-09-03
AY-6058_Caching on farm


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Caching from Maya on Deadline was feature we had in OpenPype, but because of architectural changes we dropped it.

Similar to  workflows now allowed in Houdini, we need to re-implement this in Maya too: ability to offload maya jobs to farm and publish their results. Caching (sims, fur, points, ...)

Here is original OpenPype implementation:

Related:

### Additional context:
link to discussion on [Discord]()
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-6058](https://app.clickup.com/t/86bzjgnbf)_

Prabu-7:
Hi,

Just wanted to know is this feature already deployed or still under development. I can see there was an option in ```Submit to Farm``` check option in ```pointcache``` context but it doesn't seems to so anything and publish only the workfile. Is this a feature not yet ready or i'm facing some kind of Ayon bug?

![Image](https://github.com/user-attachments/assets/478358bb-4b8b-441e-a565-01e32eb10962)

![Image](https://github.com/user-attachments/assets/c3ed2b38-c8d2-4fea-beb6-0829d35abad0)

core - 1.0.12
maya - 0.4.2
deadline - 0.5.1

m-u-r-p-h-y:
> Hi,
> 
> Just wanted to know is this feature already deployed or still under development. I can see there was an option in `Submit to Farm` check option in `pointcache` context but it doesn't seems to so anything and publish only the workfile. Is this a feature not yet ready or i'm facing some kind of Ayon bug?

> core - 1.0.12 maya - 0.4.2 deadline - 0.5.1

This issue is about fixing old OpenPype implementation of this feature and make it work in AYON again.

PR #285, opened 2025-06-06
Implementation of collect cache data

moonyuet:
## Changelog Description
This PR is to implement cache data collector for getting the relevant cache data for farm submission.
Resolve https://github.com/ynput/ayon-maya/issues/90
Related: https://github.com/ynput/ayon-deadline/pull/179

Work along with https://github.com/ynput/ayon-deadline/pull/179

## Testing notes:
1. Launch Maya
2. Publish PointCache/Animation with `Submit to Farm` option toggled on

### DIFF
--- a/client/ayon_maya/plugins/publish/collect_animation.py
+++ b/client/ayon_maya/plugins/publish/collect_animation.py
@@ -15,7 +15,7 @@ class CollectAnimationOutputGeometry(plugin.MayaInstancePlugin):
 
     """
 
-    order = pyblish.api.CollectorOrder + 0.4
+    order = pyblish.api.CollectorOrder + 0.224
     families = ["animation"]
     label = "Collect Animation Output Geometry"
 
--- a/client/ayon_maya/plugins/publish/collect_cache_farm.py
+++ b/client/ayon_maya/plugins/publish/collect_cache_farm.py
@@ -1,0 +1,33 @@
+import os
+import pyblish.api
+from ayon_maya.api import plugin
+
+
+class CollectDataforCache(plugin.MayaInstancePlugin):
+    """Collect data for caching to Deadline."""
+
+    # Run after Collect Frames
+    order = pyblish.api.CollectorOrder + 0.4991
+    families = ["publish.farm"]
+    targets = ["local", "remote"]
+    label = "Collect Data for Cache"
+
+    def process(self, instance):
+        # Should we implement this in somewhere else?
+        tmp_staging_dir = os.path.join(os.environ["AYON_WORKDIR"], "cache", "alembic")
+        os.makedirs(tmp_staging_dir, exist_ok=True)
+        # TODO: Support format other than alembic
+        filename = "{name}.abc".format(**instance.data)
+        expected_filepath = os.path.join(tmp_staging_dir, filename)
+        files = instance.data.setdefault("files", list())
+        files.append(expected_filepath)
+        expected_files = instance.data.setdefault("expectedFiles", list())
+        expected_files.append({"cache": files})
+        instance.data.update({
+            # used in MayaCacheSubmitDeadline in ayon-deadline
+            "plugin": "MayaBatch",
+            "publish": True,
+            "byFrameStep": instance.data.get(
+                "creator_attributes", {}).get(
+                    "step", 1.0)
+        })
--- a/client/ayon_maya/plugins/publish/collect_pointcache.py
+++ b/client/ayon_maya/plugins/publish/collect_pointcache.py
@@ -6,7 +6,7 @@ from maya import cmds
 class CollectPointcache(plugin.MayaInstancePlugin):
     """Collect pointcache data for instance."""
 
-    order = pyblish.api.CollectorOrder + 0.4
+    order = pyblish.api.CollectorOrder + 0.224
     families = ["pointcache"]
     label = "Collect Pointcache"
 


client/ayon_maya/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,33 @@
...
+    # Run after Collect Frames
+    order = pyblish.api.CollectorOrder + 0.4991
+    families = ["publish.farm"]
+    targets = ["local", "remote"]
+    label = "Collect Data for Cache"
+
+    def process(self, instance):
+        # Should we implement this in somewhere else?
+        tmp_staging_dir = os.path.join(os.environ["AYON_WORKDIR"], "cache", "alembic")
+        os.makedirs(tmp_staging_dir, exist_ok=True)
```
moonyuet, line 18:
```suggestion
        tmp_staging_dir = os.path.join(os.environ["AYON_WORKDIR"], "cache", "alembic")
        os.makedirs(tmp_staging_dir, exist_ok=True)
```
I know it is not preferable to do this here(it's still a bit hardcoded. But it would be great if any one can suggest better way to implement the output directory.

LiborBatek (changes requested):
Once going for ```Farm submission``` option when publishing ```pointcache``` and ```animation``` products it correctly gets to DL for rendering and once finnished it can be loaded back via loader without any issues.

There is however side effect which is imho not desirable and that is we end up with two distinct versions of ```pointcache``` / ```animation``` products  (tested both as separate jobs)

First version got published directly from ```maya``` host and the other after render on DL resulting into multiplied version (even though each having distinct ver number e.g. v045 and v046)

Im enclosing both reports for publish from maya and DL job for inspection. We should def just stick with single version depending on the submission type (local render / farm render).
[Maya-publish-report-250610-15-17.json.txt](https://github.com/user-attachments/files/20673581/Maya-publish-report-250610-15-17.json.txt)

DL job report:

[DL-Job_2025-06-10_15-17-04_684830509c110ccc489d7da5.txt](https://github.com/user-attachments/files/20673587/DL-Job_2025-06-10_15-17-04_684830509c110ccc489d7da5.txt)

LiborBatek (approved):
Now all works like charm!

I have also tested multiple characters for publishing ```animation``` products and each got exclusive DL job which is correct.

LGTM

Also during publishing the caching got skipped once ```Farm submission``` been set to ```ON``` ...again works well!

Screenshot 2025-06-10 162054

client/ayon_maya/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,19 @@
+import pyblish.api
+from ayon_maya.api import plugin
+
+
+class CollectDataforCache(plugin.MayaInstancePlugin):
+    """Collect data for caching to Deadline."""
```
BigRoy, line 6: If this targets deadline - shouldn't this be in `ayon-deadline`?

Also, this data is SO generic? Why does it need a Collector of it's own.

- `instance.data["publish"] = True` seems like a given, and even what should be default behavior?
- `"plugin": "MayaBatch"` seems awfully Deadline specific and why can't `MayaBatch` be the default plug-in in `MayaCacheSubmitDeadline`?
With that we can remove this full plug-in, no?

client/ayon_maya/plugins/publish/collect_farm_instances.py
```diff
@@ -0,0 +1,31 @@
...
+from ayon_maya.api import plugin
+
+
+class CollectFarmInstances(plugin.MayaInstancePlugin):
+    """Collect Farm Instances for remote publish
+    """
+
+    order = pyblish.api.CollectorOrder + 0.223
+    label = "Collect Farm Instances"
+    families = ["animation", "pointcache"]
```
BigRoy, line 11: Are we not supporting other product types, like `usd`, `fbx`, yeti caches and potentially ornatrix caches? Since it fully triggers logic 'locally' on the farm machine it should be trivial to support most of the regular publish product types, right? (because it behaves as if local but on the farm machine?)

client/ayon_maya/plugins/publish/collect_pointcache.py
```diff
@@ -6,14 +6,11 @@
 class CollectPointcache(plugin.MayaInstancePlugin):
     """Collect pointcache data for instance."""
 
-    order = pyblish.api.CollectorOrder + 0.4
```
BigRoy, line 9: Can this change now be reverted?

client/ayon_maya/plugins/publish/collect_animation.py
```diff
@@ -15,7 +15,7 @@ class CollectAnimationOutputGeometry(plugin.MayaInstancePlugin):
 
     """
 
-    order = pyblish.api.CollectorOrder + 0.4
```
BigRoy, line 18: Can this change now be reverted?

client/ayon_maya/plugins/publish/increment_current_file_deadline.py
```diff
@@ -13,6 +13,7 @@ class IncrementCurrentFileDeadline(plugin.MayaContextPlugin):
     order = pyblish.api.IntegratorOrder + 9.0
     families = ["workfile"]
     optional = True
+    targets = ["local", "remote"]
```
BigRoy, line 16: This seems to do the **opposite** of what we want to achieve... now the farm job will ALSO trigger this - but we don't want it to? We ONLY want to trigger this as we submit it to the farm (the UI publish session from the artist).

I believe this should only target `local`, right?

client/ayon_maya/plugins/publish/save_scene.py
```diff
@@ -12,6 +12,7 @@ class SaveCurrentScene(plugin.MayaContextPlugin):
     label = "Save current file"
     order = pyblish.api.ExtractorOrder - 0.49
     families = ["renderlayer", "workfile"]
+    targets = ["local", "remote"]
```
BigRoy, line 15: This seems to do the **opposite** of what we want to achieve... now the farm job will ALSO trigger this - but we don't want it to? We ONLY want to trigger this as we submit it to the farm (the UI publish session from the artist).

I believe this should only target `local`, right?

client/ayon_maya/plugins/publish/collect_farm_instances.py
BigRoy: Actually - no... I was stupid on my comment there. On the farm job it'd still run that full extractor.. ignore me.

client/ayon_maya/plugins/create/create_arnold_scene_source.py
```diff
@@ -47,6 +47,9 @@ def get_instance_attr_defs(self):
         defs = lib.collect_animation_defs(create_context=self.create_context)
 
         defs.extend([
+            BoolDef("farm",
+                    label="Submit to Farm",
+                    default=False),
```
BigRoy, line 52: I have my doubts about this being on all the Creators @moonyuet @philippe-ynput @iLLiCiTiT ...

This would mean this toggle would always be there, even if you have no `ayon-deadline` for example... and hence no farm? 

Plus - I may want to recommend making this an `Enum` so that if "farm" is off that the label you picked clarifies what it does then. So you'd have "Local render", "Farm render" options. (Or "Local export", "Farm export")... 

Then if for whatever reason down the line we learn that there may be another variant to this.

Also, if Farm Render is enabled... you'd suddenly also want to be able to specify the priority, pool, etc. (not sure if that's exposed currently?) - not a priority for the first version MVP but definitely something to consider down the line.

client/ayon_maya/plugins/publish/collect_farm_instances.py
```diff
@@ -0,0 +1,35 @@
...
+                "oxcache", "proxyAbc", "redshiftproxy",
+                "vrayproxy", "xgen", "yeticache"]
+    targets = ["local"]
+
+    def process(self, instance):
+        if instance.data.get("farm"):
+            instance.data["families"].append("remote_publish_on_farm")
+
+
+class CollectRemoteCacheInstances(plugin.MayaInstancePlugin):
```
iLLiCiTiT, line 21: I think this plugin does not need families filter.

At the end does not have to live in maya, but that would be bigger discussion.

client/ayon_maya/plugins/create/create_arnold_scene_source.py
BigRoy: > If it was meant for remote publish then it should be enum. Submit to farm does mot indicate remote publishing at all.

To me it's the same thing. I'm quite confused as to what "remote publish"? is exactly... it's a publish on the farm... but what it's really doing is also the Export on the farm. So it's just submitting the full export+publish to the farm. Which is essentially doing everything for the instance, on the farm. So "submit to farm" sounds accurate?

@m-u-r-p-h-y @LiborBatek what do you think artists in general would make more sense of as terminology?

LiborBatek (approved):
I have re-tested again and havent spotted any flaws or issues! Looks good!

Works fine!

client/ayon_maya/plugins/create/create_arnold_scene_source.py
philippe-ynput: @moonyuet @BigRoy "submit to farm" doesn't bother me, but it should only be enabled when available/relevant.
A `local render` / `farm render` enum would be more explicit, especially for newcomers, so I would recommend an enum.

philippe-ynput:
@moonyuet let's make sure the priority works before merging this.

moonyuet:
@philippe-ynput I have removed the priority enum here and all the priority should be set in the deadline addon instead.

LiborBatek:
@moonyuet do you plan to implement these ```exposed overrides``` within profiles for farm caching? 

```ayon+settings://deadline/publish/CollectJobInfo/profiles```

Screenshot 2025-06-18 093706

PR #289, opened 2025-06-13
Look Assigner: Fix aiStandIn assignment if gpuCache plug-in is unloaded

BigRoy:
## Changelog Description

Fix bug where `aiStandIn` nodes could not be assigned a look if `gpuCache` plug-in was not loaded.

This was due to `cmds.ls(type=["aiStandIn", "gpuCache"])` failing to return `aiStandIn` nodes if `gpuCache` node type does not exist. Now we filter out the types that do not exist.

- Introduced `nodetype_exists` function to verify if a node type is registered in the current Maya session.
- Added `arnold_standin.get_supported_node_types` to return only existing node types for look assignment.
- Refactored type hints for better clarity and consistency.

...

## Testing notes:

1. Make sure your current renderer is set to Arnold.
2. Unload `gpuCache.mll` (gpuCache plug-in)
3. Load an `aiStandIn`
4. Assign look.

It should work.

1. Load the same Alembic as `gpuCache` via loader
2. Assign look (if current renderer is arnold) should also still work.

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4440,3 +4440,23 @@ def force_shader_assignments_to_faces(shapes):
         for shading_engine, original_members in original_assignments.items():
             cmds.sets(clear=shading_engine)
             cmds.sets(original_members, forceElement=shading_engine)
+
+
+def nodetype_exists(nodetype: str):
+    """Return whether node type exists in the current Maya session.
+
+    This returns whether it's registered as a node type to maya, it does not
+    check whether it exists in the current scene.
+
+    Args:
+        nodetype (str): The node type name to check for existence.
+
+    Returns:
+        bool: True if the node type exists, False otherwise.
+    """
+    # If the node type does not exist, Maya will raise a RuntimeError.
+    try:
+        cmds.nodeType(nodetype, isTypeName=True)
+        return True
+    except RuntimeError:
+        return False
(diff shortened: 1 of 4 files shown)


moonyuet:
Is it supposed that the gpucache results to no shader assignment? 
image
otherwise looks good for aiStandin or alembic reference
image
image

BigRoy:
> Code looks good for me. I did simple test with loading pointcache product. Is it supposed that the gpucache results to no shader assignment?

No, it _should_ assign it, but only if your current renderer in the scene is set to Arnold.

LiborBatek (commented):
I have tested both scenarios and havent spotted any malfunction.

Looks good to me!

image

However there happens to be a new/another issue when ```standin``` and ```gpucache``` loaded next to each as there are issues with assignment on ```gpucache``` then...probably what @moonyuet experienced too

User been simply unable to assign ```look``` to both and it jumps from one to other when performing look assign (both ```mtoa``` and ```gpu_cache``` plugins in use) also there is no difference when using ```Assign to selected only``` and just keep malfunctioning...

there is some traceback in maya console:
```
// Warning: ayon_maya.tools.mayalookassigner.arnold_standin : Skipping setting attribute custom on ['/model_GRP/Body_GEO'] because it is not recognized.
```

BigRoy:
Thanks, that's great @moonyuet - I'd say that's an issue we should solve on load of `gpuCache` and `aiStandIn`. It's not really a look assignment issue at the core I think. But definitely something out of scope of this PR.

Aside of the Arnold namespace quirk (which was there before too) -this all works?

moonyuet:
> Thanks, that's great @moonyuet - I'd say that's an issue we should solve on load of `gpuCache` and `aiStandIn`. It's not really a look assignment issue at the core I think. But definitely something out of scope of this PR.
> 
> Aside of the Arnold namespace quirk (which was there before too) -this all works?

yes, it is. I will approve this first. But we need to make sure namespace also imported when loading ASS so that the shader won't be messed up when we assign look through look assigner.

PR #290, opened 2025-06-17
Import Loader: Relabel `Clean Import` to `Strip existing AYON asset metadata`

BigRoy:
## Changelog Description

Import Loader: Relabel `Clean Import` to `Strip existing AYON asset metadata`

Also improve the tooltip and use `BoolDef` instead of the legacy `qargparse` dependency.

## Testing notes:

1. Import option box should be clearer.
2. The option box checkbox should still work.


LiborBatek (approved):
It correctly shows the new much more descriptive and understandable label and also tested if that knob actually do what it supposed to do....all right!

No issues found.




Screenshot 2025-06-17 155823

PR #291, opened 2025-06-17
Add namespace attribute for the arnold standin when load

moonyuet:
## Changelog Description
This PR is to fix the issue found in https://github.com/ynput/ayon-maya/pull/289 when loading both gpuCache and ArnoldStandIn and using look assigner.

## Testing notes:
1. Load Ass and gpuCache assets
2. Load Look
3. Use look assigner to assign the shaders for both assets

### DIFF
--- a/client/ayon_maya/plugins/load/load_arnold_standin.py
+++ b/client/ayon_maya/plugins/load/load_arnold_standin.py
@@ -99,7 +99,7 @@ class ArnoldStandinLoader(plugin.Loader):
             cmds.setAttr(standin_shape + ".dso", path, type="string")
             sequence = is_sequence(os.listdir(os.path.dirname(repre_path)))
             cmds.setAttr(standin_shape + ".useFrameExtension", sequence)
-            cmds.setAttr(standin_shape + ".aiNamespace", namespace)
+            cmds.setAttr(standin_shape + ".aiNamespace", namespace, type="string")
 
             fps = (
                 version_attributes.get("fps") or get_fps_for_current_context()
--- a/client/ayon_maya/plugins/load/load_gpucache.py
+++ b/client/ayon_maya/plugins/load/load_gpucache.py
@@ -54,6 +54,10 @@ class GpuCacheLoader(plugin.Loader):
         path = self.filepath_from_context(context)
         cmds.setAttr(cache + '.cacheFileName', path, type="string")
         cmds.setAttr(cache + '.cacheGeomPath', "|", type="string")    # root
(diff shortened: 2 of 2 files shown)


client/ayon_maya/plugins/load/load_arnold_standin.py
```diff
@@ -99,6 +99,7 @@ def load(self, context, name, namespace, options):
             cmds.setAttr(standin_shape + ".dso", path, type="string")
             sequence = is_sequence(os.listdir(os.path.dirname(repre_path)))
             cmds.setAttr(standin_shape + ".useFrameExtension", sequence)
+            cmds.setAttr(standin_shape + ".aiNamespace", namespace)
```
BigRoy, line 102: This won't work. This will error I suspect. ;)

This needs `type="string"`, right?
```suggestion
            cmds.setAttr(standin_shape + ".aiNamespace", namespace, type="string")
```

BigRoy (changes requested):
Small change needed.

Also, we may want to do something similar on loading gpuCaches.

```python
if cmds.attributeQuery("aiNamespace", node=gpucache, exists=True):
    # Set Arnold namespace attribute to ensure shaders are loaded uniquely
    # when a gpu cache is loaded multiple times
```

However, regarding the namespace automatically. It may make sense to us now to avoid a "bug", but there is very likely a reason why Arnold doesn't automatically make the things unique. It could e.g. be that if you set unique namespaces, that each loaded standin will now populate it's own memory (and hence not be shared at rendertime as render instances?) which for e.g. a _forest_ or tons of objects may have different implications if we're enforcing unique namespaces to them. (Although one may argue that with those high amounts of objects you'd be going for a different approach perhaps than manually loading 100s of aiStandins?) Anyway, my point is... enforcing the setting of the namespace may have an adverse side effect on other scenarios.

BigRoy:
> Should we also report this to autodesk to see they would suggest a solution/or some issue track to avoid this without setting the namespace when we loaded some big forest scene?

I think getting knowledge about this would be good, yes. It's not a showstopper to get this PR through. But I'd say, let's definitely check with Autodesk or ask on their forums (+ link it here so we can find it later) and see what they say. Maybe it's absolutely fine to set `aiNamespace` by default. Only mention of `aiNamespace` google points me to is: https://forums.autodesk.com/t5/arnold-for-maya-forum/operators-using-multiple-alembic-nodes-and-override-shaders/m-p/11059418#M8245

BigRoy (approved):
image

image

Works for both. 👍

PR #293, opened 2025-06-17
Only process `dagNode` type in `get_highest_in_hierarchy`

BigRoy:
## Changelog Description

Only process `dagNode` type in `get_highest_in_hierarchy` because dependency nodes would otherwise always appear as a 'highest' object even though they are not hierarchical.

Note that we filter by `type="dagNode"` instead of `dag=True` in `cmds.ls` because `dag=True` extends the returned list to include all its children, which would return way more data then we care about and which the function would then just end up processing to directly filter out again.

Mentioned originally here: https://discord.com/channels/517362899170230292/517382145552154634/1359111623066779811

## Testing notes:

1. Publishing and loading should continue to work
2. Specifically test exporting alembics with `includeParentHierarchy` disabled.


antirotor (approved):
Looks good, publishing tested.

PR #298, opened 2025-06-23
Skip non-dag nodes in `get_all_children`

### LINKED ISSUE #297: AY-7954_Maya Rig Animation publish fails on non-dag nodes in `out_SET`

rigging publish allows various non-dag members in the out-set, like adding another object set in the out_SET  of the rig
such rig is validated and published
when such rig is used for animation, the animation product publish will fail with "no DAG node" Error.

It will error here:  

Due to the get_all_children  logic expecting only DAG nodes as inputs. Most likely it makes most sense to put a cmds.ls(nodes, type="dagNode", long=True)  at the start of get_all_children()  so that other node types are just ignored making the API function more foolproof.

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1386643201502347295)

_This issue was automatically created from Clickup ticket [AY-7954](https://app.clickup.com/t/86c4488xx)_

BigRoy:
## Changelog Description

Skip non-dag nodes in `get_all_children`

Avoids error `TypeError: item is not a DAG path`

Fix #297

## Testing notes:

1. Publish a rig with an objectSet inside the `out_SET` of the rig
2. Then with that rig loaded, try to publish animation.

It should pass and allow publishing.

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -3971,7 +3971,7 @@ def get_all_children(nodes, ignore_intermediate_objects=False):
     focus on a fast query.
 
     Args:
-        nodes (iterable): List of nodes to get children for.
+        nodes (iterable[str]): List of nodes to get children for.
         ignore_intermediate_objects (bool): Ignore any children that
             are intermediate objects.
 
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
All good...


Screenshot 2025-06-25 170442



I was able to publish 
Screenshot 2025-06-25 171022
with ```maya set``` inside of ```OUT_SET``` and load the rig and publish animation from it...


Good to go!

PR #301, opened 2025-06-24
Add create first version and keep placeholders settings to Templated Workfile Build 

BigRoy:
## Changelog Description

Add create first version and keep placeholders settings to Templated Workfile Build settings in Maya, to match e.g. `ayon-nuke` and other integrations

## Additional review information

Came up here: https://discord.com/channels/517362899170230292/563751989075378201/1387011018206609498

## Testing notes:
1. The added "create first version" and "keep placeholders" templated workfile build settings should work.

PR #307, opened 2025-06-25
Support creating first workfile with TemplateBuilder

tweak-wtf:
## Changelog Description
Supports creating the first initial workfile version from a built template.

This picks up and continues https://github.com/ynput/ayon-maya/pull/301.
I took the following commit as reference https://github.com/ynput/ayon-houdini/commit/65db6d34639e9ebf22b477bbaca9c3cd87f8c077

When testing u have to build and upload this addon as it contains server-side settings changes ⚠️

## Testing notes:
1. configure a templated workfile for maya with `create_first_version = True`
2. launch maya from a task that atm has 0 workfiles
3. check that your template gets applied accordingly
4. check the workfile is saved as initial v01 in your work-drive

*now falsify*
1. use before created template but set `create_first_version = False`
2. check that the workfile still gets applied but no workfile is saved

*and check already imported templates*
1. open up a workfile that already has your template applied
2. apply the same again
3. check script editor that it intentionally didn't do anything

### DIFF
--- a/client/ayon_maya/api/pipeline.py
+++ b/client/ayon_maya/api/pipeline.py
@@ -45,6 +45,10 @@ from ayon_core.pipeline.workfile.lock_workfile import (
     is_workfile_locked,
     is_workfile_lock_enabled
 )
+from ayon_core.pipeline.workfile.workfile_template_builder import (
+    TemplateProfileNotFound
+)
+
 from ayon_maya import MAYA_ROOT_DIR
 from ayon_maya.lib import create_workspace_mel
 
@@ -661,8 +665,14 @@ def on_new():
     with lib.suspended_refresh():
         lib.set_context_settings()
 
-    workfile_template_builder.build_workfile_template(
-        workfile_creation_enabled=True)
+    try:
+        workfile_template_builder.build_workfile_template(
+            workfile_creation_enabled=True)
+    except TemplateProfileNotFound:
+        log.debug(
+            "No workfile template profile enabled for current context. "
+            "Skipping workfile creation."
+        )
     _remove_workfile_lock()
 
 
--- a/client/ayon_maya/api/pipeline.py
+++ b/client/ayon_maya/api/pipeline.py
@@ -45,14 +45,11 @@ from ayon_core.pipeline.workfile.lock_workfile import (
     is_workfile_locked,
     is_workfile_lock_enabled
 )
-from ayon_core.pipeline.workfile.workfile_template_builder import (
-    TemplateProfileNotFound
-)
 
 from ayon_maya import MAYA_ROOT_DIR
 from ayon_maya.lib import create_workspace_mel
 
-from . import menu, lib, workfile_template_builder
+from . import menu, lib
 from .workio import (
     open_file,
     save_file,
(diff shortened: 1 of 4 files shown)


BigRoy:
> i did notice a performance degradation in launching maya.
> i suspect this is because the TemplateBuilder now gets called on any `new file` event. unsure how to circumvent this tho.

I'd say this is quite important to solve. Any ideas on how to improve best @antirotor ?

client/ayon_maya/api/pipeline.py
```diff
@@ -661,6 +661,8 @@ def on_new():
     with lib.suspended_refresh():
         lib.set_context_settings()
 
+    workfile_template_builder.build_workfile_template(
+        workfile_creation_enabled=True)
```
BigRoy, line 665: This will raise an error if no workfile template is enabled. We should avoid that.
```suggestion
    try:
        workfile_template_builder.build_workfile_template(
            workfile_creation_enabled=True
        )
    except TemplateProfileNotFound:
        # Note that debug messages are not enabled by default in Maya
        log.debug(
            "No workfile template profile enabled for current context. "
            "Skipping workfile creation."
        )
```

Requires import:
```python
from ayon_core.pipeline.workfile.workfile_template_builder import (
    TemplateProfileNotFound
)
```

BigRoy (changes requested):
This errors if no workfile templates are currently enabled for current context - other than that it looks good and seems to work as inteded.

@antirotor could you take a look at the code as well please?

client/ayon_maya/api/pipeline.py
tweak-wtf: adressed your suggestion ✌️

antirotor (changes requested):
With catching of the missing template it works, but we need to find out way not to execute the logic on startup when `AYON_LAST_WORKFILE` is set. So what about just not executing `workfile_template_builder.build_workfile_template()` if `AYON_LAST_WORKFILE` exists?

```python

...
    try:
        if not os.getenv("AYON_LAST_WORKFILE"):
            workfile_template_builder.build_workfile_template(
                workfile_creation_enabled=True)
    except TemplateProfileNotFound:
        log.debug(
            "No workfile template profile enabled for current context. "
            "Skipping workfile creation."
        )
...

```

Ideally we should not query `AYON_LAST_WORKFILE` directly but call some ayon-core function but I don't think there is one (yet) :)

BigRoy:
> however, i checked for nuke/houdini and noticed that there too the template always gets executed but existing workfiles aren't modified. it bails before modifying.

The logic there is that if when the build triggers, and the current scene is alread a saved/named file then it'll not trigger due to the assumption you have opened a workfile.

However, in Maya we have the post initialize last workfile settings (enabled by default?)

image
```
ayon+settings://maya/open_workfile_post_initialization
```

And hence likely the workfile is only opened 'later' than when the template build triggers perhaps?

moonyuet:
I tested in my side, it is successfully built the workfile by `build workfile from template` as well as `build new workfile`
image
@antirotor @BigRoy are we gonna merge this or we still need to fix something?

moonyuet (approved):

BigRoy:
@moonyuet 

This will need changes. Feel free to make a **new** PR with this as inspiration. But essentially:

- We should be able to have template builder build on app launch to create first workfile (if no workfile is being opened on launch)
- Then pressing CTRL+N (or new file) should _also_ build it again, because that's the current default behavior in AYON.
- Then we should once we have some more settings exposed be able to disable it for CTRL+N, so that it just does a regular clean file. So at some point we'll need to be able to differentiate between "new file" and "app launched".

It should work for both the case where "Open post app initialization" is disabled or enabled in ayon-maya settings.

PR #310, opened 2025-06-27
Reduce task entity and project settings queries on context resets or maya startup

BigRoy:
## Changelog Description

Re-use task entity and project settings whenever we can on resetting context attributes.

All the FPS, frame range, resolution, etc. resets on context change (save to another context) or when starting Maya should still continue to work and be faster too.

## Testing notes:

1. All the FPS, frame range, resolution, etc. resets on context change (save to another context) or when starting Maya should still continue to work and be faster too.

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -3280,7 +3280,7 @@ def update_content_on_context_change():
 
     instance_values = {
         "folderPath": create_context.get_current_folder_path(),
-        "task": create_context.get_current_task_name(),
+        "task": task_entity["name"],
     }
     creator_attribute_values = {
         "frameStart": float(task_entity["attrib"]["frameStart"]),


LiborBatek (approved):
I have tested multiple scenarios including setting context with existing publish instances and all gets attributes from the new context. Also tested on cold start on the very first start occasion.

LGTM

client/ayon_maya/api/lib.py
```diff
@@ -3249,8 +3273,8 @@ def update_content_on_context_change():
     """
 
     host = registered_host()
-    create_context = CreateContext(host)
-    task_entity = get_current_task_entity(fields={"attrib"})
+    create_context = CreateContext(host, discover_publish_plugins=False)
+    task_entity = create_context.get_current_task_entity()
```
iLLiCiTiT, line 3277: there should be `task_entity["name"]` named in `instance_values` for `"task"`.

BigRoy: Sorry, can you elaborate? How would it help in this scenario?

In this case we want to enforce all instances to match the current context.

client/ayon_maya/api/lib.py
```diff
@@ -2554,12 +2548,12 @@ def get_frame_range(include_animation_range=False):
         # Some usages of this function use the full dictionary to define
         # instance attributes for which we want to exclude the animation
         # keys. That is why these are excluded by default.
-
-        settings = get_project_settings(project_name)
+        if not project_settings:
+            project_settings = get_current_project_settings()
```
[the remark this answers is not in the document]
BigRoy: Changed with https://github.com/ynput/ayon-maya/pull/310/commits/10ff7ed29d32de9f1d460c24face96545e0853a6

client/ayon_maya/api/lib.py
BigRoy: I could use `create_context.get_current_task_name()` instead?

BigRoy: Actually, no - I couldn't - I'm using the task entity attributes further down.

iLLiCiTiT (approved):

PR #311, opened 2025-07-02
Publish: Use new workfiles API to increment current workfile.

BigRoy:
## Changelog Description

When incrementing the workfile during publish, use the new context-based workfile API so that the current author information is stored with the saved file.

Pending PR: https://github.com/ynput/ayon-core/pull/1275

## Testing notes:

1. Publishing should work
2. Workfile should be incremented
3. Workfile should show the author of the publish 
4. Workfile should have 'artist note' set: "Incremented by publishing."

### DIFF
--- a/client/ayon_maya/plugins/publish/increment_current_file_deadline.py
+++ b/client/ayon_maya/plugins/publish/increment_current_file_deadline.py
@@ -1,13 +1,14 @@
 import pyblish.api
+
+from ayon_core.lib import version_up
+from ayon_core.host import IWorkfileHost
+from ayon_core.host.interfaces import SaveWorkfileOptionalData
+from ayon_core.pipeline import registered_host
 from ayon_maya.api import plugin
 
 
 class IncrementCurrentFileDeadline(plugin.MayaContextPlugin):
-    """Increment the current file.
-
-    Saves the current maya scene with an increased version number.
-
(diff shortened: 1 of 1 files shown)


client/ayon_maya/plugins/publish/increment_current_file_deadline.py
```diff
@@ -16,25 +16,25 @@ class IncrementCurrentFileDeadline(plugin.MayaContextPlugin):
...
-        cmds.file(rename=new_filepath)
-        cmds.file(save=True, force=True, type=fileType)
+        host: IWorkfileHost = registered_host()
+        if hasattr(host, "save_workfile_with_context"):
+            from ayon_core.host.interfaces import SaveWorkfileOptionalData
+            host.save_workfile_with_context(
+                filepath=new_filepath,
+                folder_entity=context.data["folderEntity"],
+                task_entity=context.data["taskEntity"],
+                description="Incremented by publishing.",
```
BigRoy, line 29: Do we want this description set? And maybe it should be more descriptive if we do? e.g. include from which version it was incremented perhaps, or?

BigRoy:
Should be updated to match `ayon-core` changes and it's now using `save_next_version`

iLLiCiTiT (approved):

PR #312, opened 2025-07-02
Workfile: Can be set as mandatory

iLLiCiTiT:
## Changelog Description
Workfile entity can be set as mandatory.

This PR requires https://github.com/ynput/ayon-core/pull/1360 to be able to work. Is backwards compatible, it just won't do anything before that.

## Testing notes:
1. Change setting of `ayon+settings://maya/create/CreateWorkfile/is_mandatory` to `True`
2. Launch maya and open publisher.
3. Workfile instance won't have enabled checkbox and is always published.

### DIFF
--- a/client/ayon_maya/plugins/create/create_workfile.py
+++ b/client/ayon_maya/plugins/create/create_workfile.py
@@ -14,6 +14,7 @@ class CreateWorkfile(plugin.MayaCreatorBase, AutoCreator):
 
     default_variant = "Main"
 
+    settings_category = "maya"
     is_mandatory = False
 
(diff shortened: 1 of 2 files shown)


BigRoy (changes requested):
Needs only a minor tweak, see comment.

Other than that it works. It works on initial auto-create, but also if it already existed previously and was actually saved inactive - then it's still forced to active.
The ImmutableKeyError also works if you set e.g. `instance["active"] = True` after it's set to mandatory.

BigRoy (approved):

Issue #269, opened 2025-04-10
AY-7707_Maya Look Publish with Texture Reference object / load Look with proper connections made


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

We need to incorporate  Texture Reference objects into the Look product as we do not treat it atm and when the look got loaded into shot and applied on particular model/pointcache to reconnect the texRef back to the mesh to take the effect...

Should be also tested with RShift nodes besides standard Maya ones as the client uses it.

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1359814531572043846)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7707](https://app.clickup.com/t/86c30w43b)_

moonyuet:
Discussed with @LiborBatek and come with some workflow conclusion.
Things to add on Look publish:
- If we need to publish look with texture reference object, we need to export the reference object along with the scene.(We can add some options to enable taking texture reference object)
- We need to include the relevant data of the texture reference object into the look json.

Things to add on Look Assigner:
- Support to connect the texture reference object to the target object

The current look publishing workflow is publishing the scene with merely just shaders/textures. @BigRoy Do you have some advice/better battleplan on this?

PR #313, opened 2025-07-04
Support of Texture Reference Objects in Look Publish

moonyuet:
## Changelog Description
This PR is to introduce the support of Texture Reference Objects in Look Publish, users can publish their texture reference objects in the look and connect it back to the geometry with Look Assigner.

Fix https://github.com/ynput/ayon-maya/issues/269

## Testing notes:
1. Create Look with `Texture Reference Objects` options on
2. Publish
3. Load look with the related geometry
4. Use Look Assigner to assign look

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4573,7 +4573,7 @@ def apply_connections(connections, target_nodes_by_id, source_nodes_by_id):
     for connection in connections:
         source_node = next(
             iter(source_nodes_by_id.get(
-            connections["sourceID"], [])
+            connection["sourceID"], [])
             ), None
         )
 
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -2054,7 +2054,7 @@ def apply_shaders(relationships, shadernodes, nodes):
 
     # check if there is texture references input and connect the texture
     # reference back to the nodes
-    texture_references_input = relationships.get("texture_connections", [])
+    texture_references_input = relationships.get("connections", [])
     connect_texture_reference_objects(texture_references_input, nodes_by_id)
 
     # endregion
@@ -4562,8 +4562,8 @@ def connect_texture_reference_objects(texture_connections, nodes_by_id):
     """
     # Compare loaded connections to scene.
     for texture_connection in texture_connections:
-        source_node = nodes_by_id.get(input["sourceID"])
-        target_node = nodes_by_id.get(input["destinationID"])
+        source_node = nodes_by_id.get(texture_connection["sourceID"])
+        target_node = nodes_by_id.get(texture_connection["destinationID"])
 
         if not source_node or not target_node:
             self.log.debug(
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4577,7 +4577,7 @@ def apply_connections(connections, target_nodes_by_id, source_nodes_by_id):
             ), None
         )
 
-        target_nodes = target_nodes_by_id.get(connections["destinationID"], [])
+        target_nodes = target_nodes_by_id.get(connection["destinationID"], [])
 
         if not source_node or not target_nodes:
             self.log.debug(
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4601,12 +4601,8 @@ def connect_texture_reference_objects(texture_reference_inputs, nodes_by_id):
                     )
                 )
             )
-            continue
-
-        source_plug = "{}.{}".format(
-            source_node, source_attr
-        )
-        target_plug = "{}.{}".format(
+        source_plug = f"{source_node}.{source_attr}"
+        target_plug = f"{target_node}.{target_attr}"
             target_node, target_attr
         )
         if cmds.isConnected(
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4572,8 +4572,17 @@ def connect_texture_reference_objects(
     """
     # Compare loaded connections to scene.
     for texture_connection in texture_connections:
-        source_node = tx_ref_nodes_by_id.get(texture_connection["sourceID"])[0]
-        target_node = nodes_by_id.get(texture_connection["destinationID"])[0]
+        source_node = next(
+            iter(tx_ref_nodes_by_id.get(
+            texture_connection["sourceID"], [])
+            ), None
+        )
+
+        target_node = next(
+            iter(nodes_by_id.get(texture_connection["destinationID"], [])
+            ), None
+        )
+
 
         if not source_node or not target_node:
             self.log.debug(
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4558,12 +4558,12 @@ def connect_texture_reference_objects(texture_reference_inputs, nodes_by_id):
     Args:
         texture_reference_inputs (list): list of texture reference
         objects connection data
-        node_by_ids ()
+        node_by_id (dict): The dict with node ids.
     """
     # Compare loaded connections to scene.
     for reference_input in texture_reference_inputs:
-        source_node = nodes_by_ids.get(input["sourceID"])
-        target_node = nodes_by_ids.get(input["destinationID"])
+        source_node = node_by_id.get(input["sourceID"])
+        target_node = node_by_id.get(input["destinationID"])
 
         if not source_node or not target_node:
             self.log.debug(
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4558,6 +4558,7 @@ def connect_texture_reference_objects(texture_reference_inputs, nodes_by_id):
     Args:
         texture_reference_inputs (list): list of texture reference
         objects connection data
+        node_by_ids ()
     """
     # Compare loaded connections to scene.
     for reference_input in texture_reference_inputs:
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -2011,6 +2011,7 @@ def apply_shaders(relationships, shadernodes, nodes):
     shading_engines = cmds.ls(shadernodes, type="objectSet", long=True)
     assert shading_engines, "Error in retrieving objectSets from reference"
 
+    tx_ref_nodes = cmds.ls(shadernodes, type="mesh", long=True)
     # region compute lookup
     nodes_by_id = defaultdict(list)
     for node in nodes:
@@ -2019,6 +2020,11 @@ def apply_shaders(relationships, shadernodes, nodes):
     shading_engines_by_id = defaultdict(list)
     for shad in shading_engines:
         shading_engines_by_id[get_id(shad)].append(shad)
+
+    tx_ref_nodes_by_id = defaultdict(list)
+    for tx_ref_node in tx_ref_nodes:
+        tx_ref_nodes_by_id[get_id(tx_ref_node)].append(tx_ref_node)
+
     # endregion
 
     # region assign shading engines and other sets
@@ -2055,7 +2061,9 @@ def apply_shaders(relationships, shadernodes, nodes):
     # check if there is texture references input and connect the texture
     # reference back to the nodes
     texture_references_input = relationships.get("connections", [])
-    connect_texture_reference_objects(texture_references_input, nodes_by_id)
+    connect_texture_reference_objects(
+        texture_references_input, nodes_by_id, tx_ref_nodes_by_id
+    )
 
     # endregion
 
@@ -4551,7 +4559,8 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
     return linear_unit, angular_unit
 
 
-def connect_texture_reference_objects(texture_connections, nodes_by_id):
+def connect_texture_reference_objects(
+        texture_connections, nodes_by_id, tx_ref_nodes_by_id):
     """Connect texture reference object nodes to the target object
     nodes if there is one.
 
@@ -4559,11 +4568,12 @@ def connect_texture_reference_objects(texture_connections, nodes_by_id):
         texture_connections (list): list of texture reference
         objects connection data
         nodes_by_id (dict): The dict with node ids.
+        tx_ref_nodes_by_id (dict): The dict with texture reference node ids.
     """
     # Compare loaded connections to scene.
     for texture_connection in texture_connections:
-        source_node = nodes_by_id.get(texture_connection["sourceID"])
-        target_node = nodes_by_id.get(texture_connection["destinationID"])
+        source_node = tx_ref_nodes_by_id.get(texture_connection["sourceID"])[0]
+        target_node = nodes_by_id.get(texture_connection["destinationID"])[0]
 
         if not source_node or not target_node:
             self.log.debug(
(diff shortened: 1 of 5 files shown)


client/ayon_maya/plugins/publish/collect_look.py
```diff
@@ -655,6 +662,55 @@ def collect_resources(self, node):
...
+        # Store all connections
+        connections = cmds.listConnections(input_content,
+                                           source=True,
+                                           destination=False,
+                                           connections=True,
+                                           # Only allow inputs from dagNodes
+                                           # (avoid display layers, etc.)
+                                           type="dagNode",
+                                           plugs=True) or []
+        connections = cmds.ls(connections, long=True)      # Ensure long names
```
BigRoy, line 690: The instance should already have all the members (including the hierarchy collected)
```suggestion
        # Get meshes
        meshes = cmds.ls(
            list(instance),
            noIntermediate=True,
            long=True
        )
        
        # Get input connections to `referenceObject` attribute
        attrs = [f"{mesh}.referenceObject" for mesh in meshes]
        connections = cmds.listConnections(
            attrs,
            destination=False,
            connections=True,
            # Only allow input meshes
            type="mesh",
            plugs=True) or []
        )
        connections = cmds.ls(connections, long=True)      # Ensure long names
```

Note that [Maya 2023.1 added `fullNodeName` argument to `listConnections`](https://help.autodesk.com/cloudhelp/2023/ENU/Maya-Tech-Docs/CommandsPython/listConnections.html#flagfullNodeName) but we can't rely on that yet 😢

client/ayon_maya/plugins/publish/collect_look.py
```diff
@@ -655,6 +662,55 @@ def collect_resources(self, node):
...
+            dest_node, dest_attr = dest.split(".", 1)
+
+            # Ensure the source of the connection is not included in the
+            # current instance's hierarchy. If so, we ignore that connection
+            # as we will want to preserve it even over a publish.
+            if dest_attr != "referenceObject":
+                self.log.debug(
+                    "Ignoring input connection between non reference "
+                    "object nodes inside the instance: %s -> %s" % (src, dest))
+                continue
```
BigRoy, line 705: Should be resolved by above suggestion.
```suggestion
```

client/ayon_maya/plugins/publish/collect_look.py
```diff
@@ -655,6 +662,55 @@ def collect_resources(self, node):
...
+
+            inputs.append({"connections": [source_attr, dest_attr],
+                           "sourceID": lib.get_id(source_node),
+                           "destinationID": lib.get_id(dest_node)})
```
BigRoy, line 709:
```suggestion
            inputs.append({
                "connections": [source_attr, dest_attr],
                "sourceID": lib.get_id(source_node),
                "destinationID": lib.get_id(dest_node)
            })
```

client/ayon_maya/plugins/publish/extract_look.py
```diff
@@ -513,7 +514,7 @@ def process(self, instance):
                 with no_workspace_dir():
                     with lib.attribute_values(remap):
                         with lib.maintained_selection():
-                            cmds.select(sets, noExpand=True)
+                            cmds.select(sets + texture_objs, noExpand=True)
```
BigRoy, line 517: Beware: This will INCLUDE all construction history and inputs from the Texture Reference objects. If they are animated, the animation will come along. If they are constrained by something, those inputs will come along, etc.

Something to be aware of ;) not critical for now as we can explain for now to not do crazy things. But note that e.g. also the _parent hierarchy_ of the texture reference objects will come along.

client/ayon_maya/plugins/publish/collect_look.py
```diff
@@ -344,7 +344,14 @@ def collect(self, instance):
             "attributes": attributes,
             "relationships": sets
         }
-
+        if instance.data["textureRefObj"]:
+            input_collections, tx_object_nodes = (
+                self.collect_texture_reference_object_inputs(instance)
+            )
+            instance.data["lookData"].update(
+                {"reference_inputs": input_collections}
```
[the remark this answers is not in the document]
moonyuet: The latest commits would be more explicit as `texture_connections` as there is key value named `connections` inside `texture_connections` dictionary

client/ayon_maya/plugins/create/create_look.py
```diff
@@ -37,6 +37,11 @@ def get_instance_attr_defs(self):
                     label="Convert textures to .rstex",
                     tooltip="Whether to generate Redshift .rstex files for "
                             "your textures",
+                    default=self.rs_tex),
+            BoolDef("includeTextureReferenceObjects",
+                    label="Texture Reference Objects",
+                    tooltip="Whether to use texture reference objects for "
+                            "your textures",
```
BigRoy, line 44: Should this be more descriptive?
```suggestion
                    tooltip=(
                        "Whether to include texture reference objects "
                        "with the published look to reconnect to geometry "
                        "when assigning the look.",
```

client/ayon_maya/plugins/publish/collect_look.py
```diff
@@ -344,7 +344,14 @@ def collect(self, instance):
...
             "relationships": sets
         }
-
+        if instance.data["includeTextureReferenceObjects"]:
+            input_collections, tx_object_nodes = (
+                self.collect_texture_reference_object_inputs(instance)
+            )
+            instance.data["lookData"].update(
+                {"texture_connections": input_collections}
+            )
```
BigRoy, line 353: No need for `.update()` call here, plus I'd refactor `texture_connections` to just `connections`.
```suggestion
            instance.data["lookData"]["connections"] = (
                input_collections
            )
```

client/ayon_maya/api/lib.py
```diff
@@ -2011,6 +2011,7 @@ def apply_shaders(relationships, shadernodes, nodes):
     shading_engines = cmds.ls(shadernodes, type="objectSet", long=True)
     assert shading_engines, "Error in retrieving objectSets from reference"
 
+    tx_ref_nodes = cmds.ls(shadernodes, type="mesh", long=True)
```
BigRoy, line 2014: Let's rename this to `shader_mesh_nodes` so it's more accurate what it _actually_ is. Or `shader_meshes` or something that points out that these are just meshes... of the look (not of what's being assigned to.)

client/ayon_maya/api/lib.py
```diff
@@ -4544,3 +4557,85 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
+        texture_connections, nodes_by_id, tx_ref_nodes_by_id):
+    """Connect texture reference object nodes to the target object
+    nodes if there is one.
+
+    Args:
+        texture_connections (list): list of texture reference
+        objects connection data
+        nodes_by_id (dict): The dict with node ids.
+        tx_ref_nodes_by_id (dict): The dict with texture reference node ids.
+    """
```
BigRoy, line 4572: None of the logic here is _really_ specific to texture reference objects - it's just processing the connections - similar to [here](https://github.com/ynput/ayon-core/blob/b0ac87b7b12006dc8b3fffe441b91a884cc65ce5/client/ayon_core/tools/workfiles/models/workfiles.py#L360-L366).

I'd name this `apply_connections(connections, source_nodes_by_id, target_nodes_by_id)`

client/ayon_maya/api/lib.py
```diff
@@ -4544,3 +4557,85 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
+        source_node = next(
+            iter(tx_ref_nodes_by_id.get(
+            texture_connection["sourceID"], [])
+            ), None
+        )
+
+        target_node = next(
+            iter(nodes_by_id.get(texture_connection["destinationID"], [])
+            ), None
+        )
```
BigRoy, line 4584: This is wrong - I'd argue that if there are MULTIPLE targets - that each of them should be connected.

client/ayon_maya/api/lib.py
```diff
@@ -4544,3 +4557,85 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
+            self.log.debug(
+                "Could not find attribute {} on node {} for "
+                "reference input:\n{}".format(
+                    target_attr,
+                    target_node,
+                    json.dumps(
+                        texture_connection, indent=4, sort_keys=True
+                    )
+                )
+            )
```
BigRoy, line 4622: Shouldn't this `continue` or skip it then? :) Or should this continue to just error?

client/ayon_maya/plugins/publish/collect_look.py
```diff
@@ -344,7 +344,14 @@ def collect(self, instance):
...
             "relationships": sets
         }
-
+        if instance.data["includeTextureReferenceObjects"]:
+            input_collections, tx_object_nodes = (
+                self.collect_texture_reference_object_inputs(instance)
+            )
+            instance.data["lookData"]["connections"] = (
+                input_collections
+            )
```
BigRoy, line 353:
```suggestion
            connections, tx_object_nodes = (
                self.collect_texture_reference_object_inputs(instance)
            )
            instance.data["lookData"]["connections"] = (
                connections
            )
```

client/ayon_maya/plugins/publish/collect_look.py
```diff
@@ -655,6 +662,57 @@ def collect_resources(self, node):
...
+        # Store all connections
+        connections = cmds.listConnections(input_content,
+                                           source=True,
+                                           destination=False,
+                                           connections=True,
+                                           # Only allow inputs from dagNodes
+                                           # (avoid display layers, etc.)
+                                           type="mesh",
+                                           plugs=True) or []
+        connections = cmds.ls(connections, long=True)      # Ensure long names
```
BigRoy, line 690: You're better of filtering this directly to just `.referenceObject` attributes as I had pointed out here: https://github.com/ynput/ayon-maya/pull/313#discussion_r2185556770

client/ayon_maya/plugins/publish/extract_look.py
```diff
@@ -531,6 +534,10 @@ def process(self, instance):
             "attributes": lookdata["attributes"],
             "relationships": relationships
         }
+        if instance.data.get("includeTextureReferenceObjects"):
+            data.update({
+                "connections": lookdata["connections"]
+            })
```
BigRoy, line 540:
```suggestion
            data["connections"] = lookdata["connections"]
```

client/ayon_maya/plugins/publish/collect_look.py
moonyuet: I have tested the code in my side, it doesn't get any connection data for json.

moonyuet: > Works absolutely fine here? :) image
> 
> How can I reproduce this failing?

Thank you for that. I will take a look at the code again. It could be something I forgot to select in my sides.

client/ayon_maya/api/lib.py
```diff
@@ -4559,32 +4559,27 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
             texture_connection["sourceID"], [])
             ), None
         )
 
-        target_node = next(
-            iter(nodes_by_id.get(texture_connection["destinationID"], [])
-            ), None
-        )
-
+        target_nodes = nodes_by_id.get(texture_connection["destinationID"], [])
```
BigRoy, line 4580: There's something confusing here by the way :)....

`target_nodes_by_id` resolve to `source_node`? :) I think some variables are swapped, or some names are wrong somewhere.

client/ayon_maya/api/lib.py
```diff
@@ -4544,3 +4557,83 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
+
+def apply_connections(connections, target_nodes_by_id, source_nodes_by_id):
+    """Connect texture reference object nodes to the target object
+    nodes if there is one.
+
+    Args:
+        connections (list): list of texture reference
+        objects connection data
+        target_nodes_by_id (dict): The dict with target shape node ids
+        source_nodes_by_id (dict): The dict with texture reference node ids.
```
BigRoy, line 4570: This may identify type issues better.
```suggestion
        target_nodes_by_id (dict[str, list[str]]): The dict with target shape node ids
        source_nodes_by_id (dict[str, list[str]]): The dict with texture reference node ids.
```

client/ayon_maya/api/lib.py
```diff
@@ -4544,3 +4557,83 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
     angular_unit = scene_units.get("angular_units", "deg")
     return linear_unit, angular_unit
+
+
+def apply_connections(connections, target_nodes_by_id, source_nodes_by_id):
+    """Connect texture reference object nodes to the target object
+    nodes if there is one.
+
+    Args:
+        connections (list): list of texture reference
```
BigRoy, line 4567:
```suggestion
        connections (list[dict[str, Any]]): list of texture reference
```

BigRoy (changes requested):
Bug: If on pre create you enable "Texture Reference Objects" the checkbox is not enabled after create.

image
image

---

Other than that, faced some code errors - I commented those.

---

Also, it seems that there's no check whether the look nodes itself have matching ids or not with whatever you're assigning too... which, by default the texture reference object will be a 'duplicate' of the original mesh (hence they share the same `cbId` value)... which means that as soon as you're assigning looks a second time in a lighting scene then the texture reference object is already in the scene, and will be detected as an object that it should also connect to... resulting in it connecting to itself:

image

We should probably fix that the look publish (the shading nodes at least does not match ids with what it's connecting to) ;)

Other than that - it works! :)

BigRoy (changes requested):
The [issues](https://github.com/ynput/ayon-maya/pull/313#pullrequestreview-3001464970) mentioned here are still present:

- [x] Bug: If on pre create you enable "Texture Reference Objects" the checkbox is not enabled after create.
- [ ] There is no check whether the look nodes itself has matching ids or not with whatever you're assigning to. By default the texture reference object will be a 'duplicate' of the original mesh (hence they share the same cbId value)... which means that as soon as you're assigning looks a second time in a lighting scene then the texture reference object is already in the scene, and will be detected as an object that it should also connect to... resulting in it connecting to itself. We should fix that.

client/ayon_maya/api/lib.py
```diff
@@ -4579,3 +4592,89 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
+            self.log.debug(
+                "Could not find nodes for reference input:\n" +
+                json.dumps(connection, indent=4, sort_keys=True)
+            )
+            continue
+
+        # Skip if source node is one of our target shapes
+        if source_node in target_nodes:
+            log.debug("Skipping connection to source target: %s", connection)
+            continue
```
BigRoy, line 4627: It *may* solve the issue however:
- Target nodes is a `list` the lookup will be slow
- You are now excluding all the connections, not just the connection to itself... as such it'll essentially end up failing to do anything.

client/ayon_maya/api/lib.py
```diff
@@ -4579,3 +4592,89 @@ def get_scene_units_settings(project_settings=None)-> tuple[str, str]:
...
+                    source_node,
+                    json.dumps(
+                        connection, indent=4, sort_keys=True
+                    )
+                )
+            )
+            continue
+
+        for target_node in target_nodes:
+            if not cmds.attributeQuery(
```
BigRoy, line 4647:
```suggestion
            # Disallow connecting to itself if the source node
            # happens to share the same id as the node it wants
            # to target.
            if target_node == source_node:
                self.log.debug(
                    "Skipping connection to itself for {source_node} "
                    "connection {source_attr}->{target_attr}."
                )
                continue
            if not cmds.attributeQuery(
```

BigRoy (approved):
Code looks good to me. @LiborBatek can you do a test run?

PR #320, opened 2025-07-14
Layout Product Workflow: Supports to add and load transformation data of objects from the container

moonyuet:
## Changelog Description
This PR is to add an option for user to choose include transform data of every objects.
- [x] Layout Publish
- [ ] Load Layout 
n/a

## Testing notes:
1. Create Layout with objects
2. Publish Layout
3. Load Layout

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_layout.py
+++ b/client/ayon_maya/plugins/publish/extract_layout.py
@@ -94,6 +94,7 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
                 continue
 
             container_dict = self.process_containers(containers)
+            allow_obj_transforms = instance.data.get("allowObjectTransforms", False)
             for container, container_root in container_dict.items():
                 representation_id = cmds.getAttr(
                     "{}.representation".format(container))
@@ -136,7 +137,6 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
                     "extension": repre_context["ext"],
                     "host": self.hosts
                 }
-
                 local_matrix = cmds.xform(
                     container_root, query=True, matrix=True)
                 local_rotation = cmds.xform(
@@ -168,6 +168,15 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
                     "y": local_rotation[1],
                     "z": local_rotation[2]
                 }
+                if allow_obj_transforms:
+                    container_objects = cmds.listRelatives(
+                        container_root, allDescendents=True, type='transform'
+                    ) or []
+                    for container_object in container_objects:
+                        json_element = (
+                            self.parse_objects_transform_as_json_element(
+                                container_object, json_element)
+                        )
                 json_data.append(json_element)
         json_filename = "{}.json".format(instance.name)
         json_path = os.path.join(stagingdir, json_filename)
@@ -245,3 +254,35 @@ class ExtractLayout(plugin.MayaExtractorPlugin):
         convert_transform.setScale([convert_scale[0], convert_scale[2], convert_scale[1]], om.MSpace.kWorld)
 
         return convert_transform.asMatrix()
+
+    def parse_objects_transform_as_json_element(self, container_object, json_element):
+        """Parse transform data of the container objects and add it as json element
+
+        Args:
+            container_object (str): Objects inside the container
+            json_element (dict): Json element
+
+        Returns:
+            dict: json element with the added transform data of the container object
+        """
+        if container_object:
+            if container_object not in json_element:
+                json_element[container_object] = {}
+            local_matrix = cmds.xform(
+                container_object, query=True, matrix=True)
+            local_rotation = cmds.xform(
+                container_object, query=True, rotation=True, euler=True)
+
+            t_matrix = self.create_transformation_matrix(local_matrix, local_rotation)
+
+            json_element[container_object]["transform_matrix"] = [
+                list(row)
+                for row in t_matrix
+            ]
+            json_element[container_object]["rotation"] = {
+                "x": local_rotation[0],
+                "y": local_rotation[1],
+                "z": local_rotation[2]
+            }
+
+        return json_element
--- a/client/ayon_maya/plugins/create/create_layout.py
+++ b/client/ayon_maya/plugins/create/create_layout.py
@@ -17,5 +17,10 @@ class CreateLayout(plugin.MayaCreator):
                     label="Group Loaded Assets",
                     tooltip="Enable this when you want to publish group of "
                             "loaded asset",
-                    default=False)
+                    default=False),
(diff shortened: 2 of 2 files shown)


client/ayon_maya/plugins/load/load_layout.py
```diff
@@ -148,6 +148,16 @@ def set_transformation(self, assets, element):
             # flatten matrix to a list
             maya_transform_matrix = [element for row in transform for element in row]
             self._set_transformation_by_matrix(asset, maya_transform_matrix)
+            for obj in element.get("object_transform", []):
+                obj_transform_name = obj["name"]
+                obj_transforms = cmds.ls(obj_transform_name, transforms=True)
```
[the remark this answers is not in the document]
moonyuet: It would be returning the object's name with the namespace, please suggest some ways to ensure the selection of the loaded data. Thank you!

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -168,6 +172,15 @@ def process(self, instance):
...
                 }
+                if allow_obj_transforms:
+                    child_transforms = cmds.listRelatives(
+                        get_all_children([container_root]), type="transform"
+                    )
+                    for child_transform in child_transforms:
+                        json_element = (
+                            self.parse_objects_transform_as_json_element(
+                                child_transform, json_element)
+                        )
```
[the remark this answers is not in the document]
BigRoy: Provided code suggestions just now - it should be a trivial change really.

- https://github.com/ynput/ayon-maya/pull/320#discussion_r2218705304
- https://github.com/ynput/ayon-maya/pull/320#discussion_r2218708120

client/ayon_maya/plugins/load/load_layout.py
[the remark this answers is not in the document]
moonyuet: I tried with the same objects with different namespaces, it works.   
image
image

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -168,6 +172,22 @@ def process(self, instance):
                     "y": local_rotation[1],
                     "z": local_rotation[2]
                 }
+                if allow_obj_transforms:
+                    child_transforms = cmds.listRelatives(
```
BigRoy, line 176: You seem to have reverted this to `listRelatives` for some reason. This is wrong.
- `get_all_children` already gets the child relatives.
- Also, you're lacking the `fullPath=True`.  But it should be a `cmds.ls()` call with `long=True` actually - which is what it was before. Not sure why it was reverted?

client/ayon_maya/plugins/load/load_layout.py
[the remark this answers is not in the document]
BigRoy: It works, as in - it by luck gets the right name? ;) If I load the same layout three times, and update them in random orders - does it always correctly update and set the correct ones?

client/ayon_maya/plugins/publish/extract_layout.py
moonyuet: You can test it by your own. If it is cmds.ls() by the case, it would mess up the entire transformation data and once you load from the layout loader, it won't replicate the same scene as your workfile has been exported.

BigRoy: Here's an example:
image

image

client/ayon_maya/plugins/publish/extract_layout.py
```diff
@@ -168,6 +172,23 @@ def process(self, instance):
...
+                        long=True
+                    )
+                    if child_transforms:
+                        object_transforms = json_element.setdefault("object_transform", [])
+                        for child_transform in child_transforms:
+                            object_transforms.append(
+                                self.parse_objects_transform_as_json_element(
+                                    child_transform
+                                )
+                            )
```
BigRoy, line 191: I assume the `child_transform` 'name' should be unique inside the data, right? Like a `name` should never occur twice when extracting here - they need to be unique by design.

If so, I'd restructure the data to use a `dict` here instead of `list` and make the key the object name.

So that it becomes:
```python
{
    "object1": {"transform_matrix": t_matrix1},
    "object2": {"transform_matrix": t_matrix2}
}
```

And, since `euler` values are unused and if we do not need to store additional data we can even just make this:
```python
{
    "object1": t_matrix1,
    "object2": t_matrix2
}
```
Greatly simplifying the data structure?

BigRoy: Is that the "rotation" aspect? I'm pretty sure if that's the case something is lacking or broken, because with just `euler` values you're lacking the rotation order information.

This object has the same rotation euler values, but differs in rotation order.

image

```
xform -query -rotation -euler;
// Result: 75.771793 -64.304802 -85.400775
select -r pCube2 ;
xform -query -rotation -euler;
// Result: 75.771793 -64.304802 -85.400775
```

So whatever we're currently doing with these euler values - it'll be inaccurate and incomplete without also knowing the rotation order.

The only reason I can see for storing euler rotation is trying to preserve the <-360 and >360 degree euler angles because the transformation matrix won't contain those. However, for euler angle information to be complete, you'd need to know the rotation order?

Anyway, feel free to keep the rotation values if there's a use case for it. In that case just change it to `dict`.

> are we ignoring the unreal aspect of it? This extractor is not only for maya but also for unreal too.

@antirotor what do you think?

moonyuet: > select -r pCube1
> xform -query -rotation -euler;
> // Result: 75.771793 -64.304802 -85.400775
> select -r pCube2 ;
> xform -query -rotation -euler;
> // Result: 75.771793 -64.304802 -85.400775
> ```

We can use quaternion to get the rotation data but this also means we need to make changes on not only the rotation data in maya but also in unreal too.

BigRoy: > This is for loading in unreal layout only, remember how the alembic orientation messes up on the unreal and they are not correct? I would comment on this if this sounds confusing to any users.

I understand some workaround was implemented - but this just boils back down to the fact that we've worked around something we couldn't figure out. The transformation matrix completely describes the objects transformation, including its rotation. We shouldn't be needing the euler angles because we can completely extract that information from the matrix (up to -180/+180 degrees, and beyond that it'd flip but remain the same rotation). Anyway, my point is that the current workaround _must_ be broken because we're lacking rotation order details, whereas the transformation matrix _should_ completely describe the object's translate, rotate, scale and shear even if that means we'd need to 'reorient' it to up/forward-axis etc inside another DCC which is where I think we've failed all this time?

Anyway, not an issue generated by this PR so let's avoid that scope sneaking in here.

PR #322, opened 2025-07-15
Creator Settings: Add settings for all creators

### LINKED ISSUE #306: The names of creator python class and server settings do not match.

## Current Behavior

The creator python class name is "CreateRenderlayer".
C:\Users\<user>\AppData\Local\Ynput\AYON\addons\maya_0.4.9\ayon_maya\plugins\create\create_render.py
```python
class CreateRenderlayer( plugin.RenderlayerCreator ):
```

But the corresponding sever settings is "CreateRender".
ayon-maya\server\settings\creators.py
```python
CreateRender: BasicCreatorModel = SettingsField(
    default_factory=BasicCreatorModel,
    title="Create Render"
)
```

Unmatched naming leading to that this creator could could be turn off or setup.

And missing server settings for these creators:

CreateMatchmove
CreateMayaUsdLayer
CreateMayaUsd
CreateMultishotLayout
CreateRedshiftProxy
CreateUnrealYetiCache
CreateXgen
CreateYetiCache

Image

### Steps To Reproduce:

_No response_

### Version
server 1.10.0
maya 0.4.9

### Relevant log output:

```shell

```

BigRoy:
## Changelog Description

Expose and fix all Creator settings. You can now freely disable all creators if you wish :)

Creator settings are now also ordered alphabetically.

Fix #306

## Testing notes:

1. All creator settings should work.

### DIFF
--- a/client/ayon_maya/plugins/create/create_render.py
+++ b/client/ayon_maya/plugins/create/create_render.py
@@ -54,9 +54,7 @@ class CreateRender(plugin.RenderlayerCreator):
         if self.render_settings.get("apply_render_settings"):
             lib_rendersettings.RenderSettings().set_default_renderer_settings()
 
-        super(CreateRenderlayer, self).create(product_name,
-                                              instance_data,
-                                              pre_create_data)
+        super().create(product_name, instance_data, pre_create_data)
 
     def get_instance_attr_defs(self):
         """Create instance settings."""
--- a/client/ayon_maya/plugins/create/create_render.py
+++ b/client/ayon_maya/plugins/create/create_render.py
@@ -12,7 +12,7 @@ from ayon_core.lib import (
 )
 
 
-class CreateRenderlayer(plugin.RenderlayerCreator):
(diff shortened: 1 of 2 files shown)


client/ayon_maya/plugins/create/create_render.py
```diff
@@ -12,7 +12,7 @@
 )
 
 
-class CreateRenderlayer(plugin.RenderlayerCreator):
+class CreateRender(plugin.RenderlayerCreator):
```
moonyuet, line 15: Do we need to implement the backward compatibility for `CreateRenderLayer` or?

BigRoy: As far as I know, we don't. We either change the class name or set `settings_name` attribute. But the class name isn't stored or imprinted anywhere as far as I know, we always refer to the creator identifiers instead so the class name should be safe to rename.

moonyuet (approved):
Looks good. all the plugins are at the alphabetical order
image

PR #329, opened 2025-08-01
Bug: remove broken method creating connections on load

antirotor:
## Changelog Description
Fix loading issue where unnecessary connections were made on load of particular references. This caused even thousands of connections to be created immensely slowing down the scene on load.

Original purpose of this function was to force load containers from the loaded reference to main container but because of the reversed condition it didn't work. Instead, it took all nodes with `.id` attributes, moved it to set and that created those connections.

## Testing notes:
Loading should still work fine, Loading of larger product with nodes having any `.id` attribute shouldn't create any unnecessary connections.


LiborBatek (approved):
Been tested in ```maya 2026``` and all working nicely...

I cant find any unnecessary set nodes connection even when re-opening the workfile.

LGTM

image

PR #333, opened 2025-08-14
Supports to auto create camera instance

moonyuet:
## Changelog Description
Brief description of the solution this PR is implementing. This should be only a few sentences and it will be publicly visible in the changelog. You can highlight importance of the change and add links to further information or supporting documentation

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -90,6 +90,15 @@ class RigSetsNotExistError(RuntimeError):
     pass
 
 
+class CameraInstanceCreationError(RuntimeError):
+    """Raised when camera instance creation fails.
+
+    This is raised from `create_camera_instance` when the required
+    camera nodes are missing or invalid.
+    """
+    pass
+
+
 def get_main_window():
     """Acquire Maya's main window"""
     from qtpy import QtWidgets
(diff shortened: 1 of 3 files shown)


client/ayon_maya/api/lib.py
```diff
@@ -4377,6 +4386,68 @@ def create_rig_animation_instance(
...
+    Args:
+        nodes (list): The loaded camera nodes.
+        context (dict): Representation context of the camera container.
+        namespace (str): The namespace for the new camera instance.
+        options (dict, optional): Additional options for the camera instance.
+        log (logging.Logger, optional): Logger to log to if provided.
+    """
+    if options is None:
+        options = {}
+    print(f"nodes: {nodes}")
```
BigRoy:
```suggestion
```

BigRoy:
> ok, but now it even happens when e.g. loading Camera product which to me is not desired...as its basically already published camera animation. It should only happen when Camerarig been brought into active workfile similar to rig and not already published animation product

Ah, valid point - I wonder if it makes any sense to allow the settings to specify the product types, so that one _could_ technically even enable this for say loading a `mayaScene` too. But I always forget about `cameraRig` product type existing. Anyway, I'm now also unsure whether having it _also_ happen on loading any camera into the scene, whether that should auto-produce a publishable instance as well. Hmmm... Will leave that to your preference to decide @LiborBatek (or perhaps the client).

LiborBatek (changes requested):
Gives me this error while trying to load a `CameraRig` product into active workfile

```
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.6.6\ayon_core\tools\loader\models\actions.py", line 747, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.6.6\ayon_core\pipeline\load\utils.py", line 324, in load_with_repre_context
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.6.6\ayon_core\pipeline\load\plugins.py", line 419, in wrapped_method
    result = original_method(self, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.4.17+dev\ayon_maya\api\plugin.py", line 812, in load
    self.process_reference(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.4.17+dev\ayon_maya\plugins\load\load_reference.py", line 247, in process_reference
    self._post_process_camera(namespace, context, options)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.4.17+dev\ayon_maya\plugins\load\load_reference.py", line 337, in _post_process_camera
    create_camera_instance(
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.4.17+dev\ayon_maya\api\lib.py", line 4410, in create_camera_instance
    assert camera_nodes, "No camera nodes in camera, this is a bug."
           ^^^^^^^^^^^^
AssertionError: No camera nodes in camera, this is a bug.
```

LiborBatek:
I was able to produce the `camerarig` and loaded and it successfully created the `camera` instance within the active workfile...
there are few flaws still...

1)First the `camerarig` subset could have just simpler naming without any prefixes just e.g. `cam` or `camera` good example been `rig` with its subsets...

Screenshot 2025-11-06 163156

2) when `camerarig` loaded the publish instance for `camera` is succesfully created but the Variant and resulting name is kinda off (too loong and unclear) we should really strip it to just `Main` as when creating the publish instance manually



Screenshot 2025-11-06 163421

so rather like this:

Screenshot 2025-11-06 163438


And lastly it fails when trying to publish such product `camera` with following error:


Screenshot 2025-11-06 163624

client/ayon_maya/api/lib.py
```diff
@@ -4377,6 +4386,66 @@ def create_rig_animation_instance(
...
+        log (logging.Logger, optional): Logger to log to if provided.
+    """
+    if options is None:
+        options = {}
+
+    camera_nodes = [
+        node for node in nodes if cmds.referenceQuery(
+        node, isNodeReferenced=True) and
+        cmds.listRelatives(node, children=True, type="camera")
+    ]
```
BigRoy: This will be relatively slow because you're doing `listRelatives` call for each node individually.

Also, `listRelatives` suffers from the fact that without `fullPath=True` it by defaults does not return unique paths if there are objects with same names in the scene.

And note that `listRelatives` returns `None` if there are no valid things to return... (odd API in maya I know). We are only passing it to `cmds.select` which takes `None` just fine, regardless I'd prefer the type to be consistent so use `or []`

So I'd do:
```suggestion
    referenced_nodes: list[str] = [
        node for node in nodes 
        if cmds.referenceQuery(node, isNodeReferenced=True)
    ]
    camera_nodes: list[str] = cmds.listRelatives(
        referenced_nodes, children=True, type="camera", fullPath=True
    ) or []
```

Issue #280, opened 2025-05-23
Arnold: Multipart set on any AOV driver will mark rest of them multipart


antirotor:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
When one AOV is set as multipart, AYON will mark all other as mutlipart too. This will result in reviews being created for all of them.

## Expected Behavior
Only products with multipart should be processed with "multipart" review logic -"simple" exrs are marked for review based on their name regex (searching for "beauty' by default). If multipart, review is set and determined by channels.

### Steps To Reproduce:

- Create multiple render layers with AOVs
- Create some light groups and add their AOVs.
- Override `multipart` attribute on one of them (have it off for the rest, remember to use `<RenderPass>` token)
- Run validation - in the output of `Collect Render Layers` you should be able to see what is flagged as multipart.

### Version
ayon-maya 0.4.7 - Maya 2026

### What platform were you running when you found the bug?
Windows

### Relevant log output:

```shell

```

PR #334, opened 2025-08-18
Arnold: Multipart should be collected correctly for each AOV driver

moonyuet:
## Changelog Description
This PR is to make sure collecting Arnold multipart information for AOVs based on their AOV drivers, but not the default one.
Resolve https://github.com/ynput/ayon-maya/issues/280

n/a

## Testing notes:
1. Create multiple renderlayers with different AOVs
2. These AOVs are hooked with different AOV drivers
3. Some of the drivers are clicked without multipart and mergeAovs, while some are with these and some are with Light Groups.
4. Publish
5. Multipart should be correct.

### DIFF
--- a/server/settings/render_settings.py
+++ b/server/settings/render_settings.py
@@ -487,7 +487,7 @@ DEFAULT_RENDER_SETTINGS = {
     "arnold_renderer": {
         "image_prefix": "<Scene>/<RenderLayer>/<RenderLayer>_<RenderPass>",
         "image_format": "exr",
-        "multilayer_exr": False,
+        "multilayer_exr": True,
         "tiled": True,
         "aov_list": [],
         "additional_options": []


moonyuet:
> Sorry but I really struggle to setup a maya scene which works for testing....not sure whats going wrong but suddenly I cant produce my AOVs (simply not rendering into framebuffer).
> 
> Im also not fully sure about the output driver and its settings...as its not clear and quite confusing. So should I setup defaultOutputDriver to `mergeAOV` `Enabled` ? as it does seem not but `multipart` working only when layers / AOVs been merged together into single file.
> 
> Also the testing steps are pretty vaque even trying different scenarios I have ended up with failure on DL publish job (wrong naming on expected files etc.) Its a combination of maya lightgroup setup mystery plus Ayon magic among others...
> 
> Maybe adding step by step might help, sorry.

the multilayer is True when either `multipart` or `merge AOVs` in the AOV drivers are enabled, as well as all light groups option is triggered.

You can test the PR by creating new AOV drivers for particular AOVs

https://github.com/user-attachments/assets/5ff90486-fa82-4ae0-9318-8ff4fe4cf06b

LiborBatek:
@moonyuet 
Thanks for your video....even tho I know how to hook a new ```ArnoldOutputDriver``` and maybe easier way...I also get that, one can change those ```Multipart``` and ```MergeAovs``` settings per each ArnoldOutputDriver but...

what should be my settings in ```Render Globals``` ??
Screenshot 2025-08-19 133632

as Ondrej stated in the ```Issue``` that user should use token ```<RenderPass>``` which can be used only if non-merged Aov approach been used. As with ```MergeAOVs``` Ayon validators will fail on wrong naming format due to "RenderPass" token present...if MergeAOVs enabled. 

I suppose we should test with mergedAOVs aka multilayer EXR not single layer, right? Correct me pls if wrong assumption.

And lastly, what role plays ```LightGroups``` as these caused me a lot of headache during testing in maya generally....could I just leave those out and test with just native available AOVs of maya with custom created ArnoldOutputDriver for each? 

It would be then much easier to test...having answers to these.

Thanks for clarification!

client/ayon_maya/plugins/publish/collect_render.py
```diff
@@ -118,7 +118,7 @@ def process(self, instance):
         for product in render_products:
             self.log.debug(f"Getting render product: {product}")
             if product.multipart:
-                multipart = True
+                multipart = product.multipart
```
BigRoy, line 121: Does this really change anything, since it's still in the `if` check, or?

LiborBatek (approved):
I have configured multiple AOVs and different Arnold output drivers and configured each with own settings (multipart enabled or disabled)

By checking my rendered files / publish report I can tell they differ inside following these preferences...

Looks ok to me. I havent touched any lightgroups and AOVs for simplicity as I had issues not connected to this PR in maya...

Here is a proof my output drivers were correctly resolved each:
Screenshot 2025-08-19 161932

NOTE:
originally I had in mind I could make renderLayer override for each AOVs and its OutputDriver (overriding each on different renderLayer) but I couldnt handle such override as maya denied it and reported errors when doing such override so I left it without any and rather introduced new Output driver for particular AOV instead.

I hope I have tested well...

client/ayon_maya/plugins/publish/validate_rendersettings.py
```diff
@@ -237,7 +237,10 @@ def get_invalid(cls, instance):
                     dir_prefix))
 
         if renderer == "arnold":
-            multipart = cmds.getAttr("defaultArnoldDriver.mergeAOVs")
+            multipart = (
+                cmds.getAttr("defaultArnoldDriver.mergeAOVs") or
+                cmds.getAttr("defaultArnoldDriver.multipart")
```
BigRoy, line 242: I vaguely recall enabling multipart didn't actually _enforce_ AOVs to be merged - it would just be 'irrelevant' if enabled but not merging parts? Or? Hence, multipart being enabled doesn't actually enforce it results in a single file which as far as I know is what we use the `multipart` variable for? Hence, the question - is this change actually correct?

LiborBatek: afaik yes,  Multipart  is not tied to be used only with multilayered EXR and been completely separate format (neither supported in single layered EXR nor multilayered one) so probably not needed...

client/ayon_maya/plugins/publish/collect_render.py
moonyuet: It is supposed that renderProduct has stored multipart as boolean value. But to ensure there is any nonetype-value bug, I would use the boolean value as if before.

client/ayon_maya/plugins/publish/collect_render.py
```diff
@@ -314,6 +314,7 @@ def process(self, instance):
 
         # Update the instance
         instance.data.update(data)
+        self.log.debug(f"multipart: {data.get('multipart')}")
```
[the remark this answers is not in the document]
BigRoy: Did it actually log anything correctly? ;) since I don't see where it's stored to `data["multipart"]`?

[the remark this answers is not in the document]
moonyuet: > Did it actually log anything correctly? ;) since I don't see where it's stored to `data["multipart"]`?
> 
> Hmm, it did list `True` for Libor's screenshot. So somehow we're storing it to `data['multipart"]` and `data["multipartExr"]`?

Ummm I can tell it it does nothing. Getting render product debug did the print:
https://github.com/ynput/ayon-maya/blob/4b95d59f4c5fb34da77b62a8bd8f7b6258a38b06/client/ayon_maya/plugins/publish/collect_render.py#L119

Issue #131, opened 2024-10-01
AY-2219_Lock rig parent objectSet


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

So we had an issue with an animator loading the published rig into the shot to animate and having deleted the rigMain parent set.  This obviously breaks the publishing for animation. This could be prevented by Locking the set right after creation.  I think this is an important one to catch. (upraveno)

### Additional context:
()
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-2219](https://app.clickup.com/t/86bw42j1t)_

PR #340, opened 2025-09-05
Add options of lock set on load when loading rig for animation publish

moonyuet:
## Changelog Description
This PR is to add lock set on load in ayon setting so that users load the rig for animation publish. This would prevent users accidentally delete the instance during animating process.
Resolve https://github.com/ynput/ayon-maya/issues/131

n/a

## Testing notes:
1. Enable `ayon+settings://maya/load/reference_loader/lock_set_on_load` and disable it 
image

2. Load Rig
3. Animating it
4. Publish
5. Should be all working
6. Removing instance is also working as expected

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -109,8 +109,10 @@ def unlocked(node):
     Args:
         node (str): The name of the node to unlock.
     """
-    has_locked = cmds.lockNode(node, query=True, lock=True)
-    cmds.lockNode(node, lock=False)
+    has_locked = False
+    if cmds.lockNode(node, query=True, lock=True):
+        has_locked = True
+        cmds.lockNode(node, lock=False)
 
     try:
         yield
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -109,10 +109,8 @@ def unlocked(node):
     Args:
         node (str): The name of the node to unlock.
     """
-    has_locked = False
-    if cmds.lockNode(node, query=True, lock=True):
-        has_locked = True
-        cmds.lockNode(node, lock=False)
+    has_locked = cmds.lockNode(node, query=True, lock=True)
+    cmds.lockNode(node, lock=False)
 
     try:
         yield
@@ -4340,21 +4338,6 @@ def create_rig_animation_instance(
                 "lock_instance": options.get("lock_instance", False)
             }
         )
-
-
-def is_animation_instance(objectset: str) -> bool:
-    """Check if the given object set is an animation instance.
-
-    Arguments:
-        objectset (str): The name of the object set to check.
-
-    Returns:
-        bool: True if the object set is an animation instance, False otherwise.
-    """
-    creator_identifier = cmds.getAttr(f"{objectset}.creator_identifier")
-    if creator_identifier == "io.openpype.creators.maya.animation":
-        return True
-    return False
 
 
 def get_node_index_under_parent(node: str) -> int:
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4262,20 +4262,6 @@ def get_creator_identifier(node: str) -> str | None:
     return get_attribute(f"{node}.creator_identifier")
 
 
-def is_animation_instance(objectset: str) -> bool:
-    """Check if the given object set is an animation instance.
-
-    Arguments:
-        objectset (str): The name of the object set to check.
-
-    Returns:
-        bool: True if the object set is an animation instance, False otherwise.
-    """
-    return (
-        get_creator_identifier(objectset) == "io.openpype.creators.maya.animation"
-    )
-
-
 def create_rig_animation_instance(
     nodes, context, namespace, options=None, log=None
 ):
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -103,6 +103,29 @@ def get_main_window():
 
 
 @contextlib.contextmanager
+def unlocked_node(node, product_type):
+    """Unlock a node for the duration of the context.
+
+    Args:
+        node (str): The name of the node to unlock.
+        product_type (str): The type of product being processed.
+    """
+    if product_type != "animation":
+        yield
+        return
+
+    has_locked = False
+    if cmds.lockNode(node, query=True, lock=True):
+        has_locked = True
+        cmds.lockNode(node, lock=False)
+
+    try:
+        yield
+    finally:
+        cmds.lockNode(node, lock=has_locked)
+
+
+@contextlib.contextmanager
 def suspended_refresh(suspend=True):
     """Suspend viewport refreshes
 
@@ -4231,7 +4254,7 @@ def get_reference_node_parents(ref):
 
 def create_rig_animation_instance(
     nodes, context, namespace, options=None, log=None,
-    lock_set_on_load=False
+    settings=None
 ):
     """Create an animation publish instance for loaded rigs.
 
@@ -4244,7 +4267,7 @@ def create_rig_animation_instance(
         namespace (str): Namespace of the rig container
         options (dict, optional): Additional loader data
         log (logging.Logger, optional): Logger to log to if provided
-        lock_set_on_load (bool, optional): Whether to lock the set on load
+        settings (dict, optional): Loader settings from Ayon
 
     Returns:
         None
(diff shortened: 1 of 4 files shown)


client/ayon_maya/api/lib.py
```diff
@@ -4320,6 +4322,8 @@ def create_rig_animation_instance(
             pre_create_data={"use_selection": True}
         )
 
+    cmds.lockNode(f"*{namespace}", lock=lock_set_on_load)
```
BigRoy, line 4325: This does more than lock the set. This has a high change of matching way more than intended. We should **explicitly** target the one set you want to lock to do this correctly.

client/ayon_maya/api/plugin.py
```diff
@@ -992,6 +992,11 @@ def remove(self, container):
         """
         from maya import cmds
 
+        locknode_namespace = container["namespace"]
+        # check if the node has been locked and unlocked it
+        if cmds.lockNode(f"*{locknode_namespace}", query=True, lock=True):
+            cmds.lockNode(f"*{locknode_namespace}", lock=False)
```
BigRoy, line 998: I'm so confused about this - **what** set are we referring to here? The animation instance? Why can't we be more accurate and **actually** find the instance correctly instead of wildcarding with a name (that is actually intended  to indicate a namespace?)

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -215,7 +215,15 @@ def process_reference(self, context, name, namespace, options):
...
-                self._post_process_rig(namespace, context, options)
+                lock_set_on_load = settings['maya']['load'].get(
+                    'reference_loader', {}
+                ).get('lock_set_on_load', False)
+                self._post_process_rig(
+                    namespace,
+                    context,
+                    options,
+                    lock_set_on_load=lock_set_on_load
+                )
```
BigRoy, line 226: Why not just pass the `settings`? That would also allow more data to flow through down the line?

server/settings/loaders.py
```diff
@@ -110,6 +110,9 @@ class ReferenceLoaderModel(BaseSettingsModel):
     display_handle: bool = SettingsField(
         title="Display Handle On Load References"
     )
+    lock_set_on_load: bool = SettingsField(
+        title="Lock Set on Load"
+    )
```
BigRoy, line 115: This needs to be way more descriptive. Lock set, why? Which set?
Also this only applies to rigs? That **must** be clear then because it's on the ReferenceLoader model.

BigRoy (changes requested):

client/ayon_maya/api/plugin.py
```diff
@@ -993,11 +996,17 @@ def remove(self, container):
         from maya import cmds
 
         node = container["objectName"]
-
+        namespace = container["namespace"]
+        associated_namespace_node = next((
+            associated_node for associated_node
+            in cmds.ls(f"*{namespace}", type="objectSet")),
+            None
+        )
```
BigRoy, line 1004: I still don't think it's the right approach to just assume that anything ending with that namespace name will be the object set you want to delete. This has a chance of capturing the WRONG associated node which we absolutely never want to have happen.

Also, this seems very specific to the rig set - even though the logic itself does not HINT that it is nor provide any comments about it.

It might be more explicit to move this call to a separate method,

```
if product_type == "rig":
    self._remove_linked_animation_instance(container)
```

Then in the method make sure to accurately find the animation instance.
As such you may need to find ANY object sets that contain ANY of the nodes from this container, and if that object set is an publish instance of the animation identifier, then remove it. (for which I feel you _should_ actually be calling the relevant remove instance call on the Create Context)

server/settings/loaders.py
```diff
@@ -110,6 +110,9 @@ class ReferenceLoaderModel(BaseSettingsModel):
     display_handle: bool = SettingsField(
         title="Display Handle On Load References"
     )
+    lock_set_on_load: bool = SettingsField(
+        title="Lock Set when Create Animation Instance on load"
+    )
```
BigRoy, line 115: Please use the description or comment argument and make this **more** descriptive. The majoity of users do not KNOW when an animation instance is created - also, in a way the statement is not true. It does not actually do so when creating the animation instance, it does so when loading a rig which happens to lock the animation instance. 

So in essence it's:
- Lock animation instance on rig load.

Right?

Assume that someone browsing the settings has little knowledge of what settings are available and how everything works. Guide them along. Especially for this setting that is in a ReferenceLoader section for which it feels quite oddly positioned when loading rigs.

We'll want to also give them something that hopefully may catch them when trying to search for it, e.g. when searching "rig", "animation", or "lock". etc.

client/ayon_maya/api/lib.py
```diff
@@ -102,6 +102,24 @@ def get_main_window():
...
+def unlocked(node):
+    """Unlock a node for the duration of the context.
+
+    Args:
+        node (str): The name of the node to unlock.
+    """
+    has_locked = False
+    if cmds.lockNode(node, query=True, lock=True):
+        has_locked = True
+        cmds.lockNode(node, lock=False)
```
BigRoy, line 115:
```suggestion
    has_locked = cmds.lockNode(node, query=True, lock=True)
    cmds.lockNode(node, lock=False)
```

client/ayon_maya/api/lib.py
```diff
@@ -4317,10 +4335,28 @@ def create_rig_animation_instance(
...
+    Arguments:
+        objectset (str): The name of the object set to check.
+
+    Returns:
+        bool: True if the object set is an animation instance, False otherwise.
+    """
+    creator_identifier = cmds.getAttr(f"{objectset}.creator_identifier")
+    if creator_identifier == "io.openpype.creators.maya.animation":
+        return True
+    return False
```
BigRoy, line 4357: I feel like this shouldn't be in `lib.py` - it's quite specific honestly to `load_reference.py` logic. Likely best moved there. This isn't generic enough to be in lib.py.

client/ayon_maya/api/plugin.py
```diff
@@ -1028,10 +1022,20 @@ def remove(self, container):
...
         except RuntimeError:
             pass
 
+    def _remove_linked_animation_instance(self, container):
+        namespace = container["namespace"]
+        node = next((
+            node for node
+            in cmds.ls(f"*{namespace}", type="objectSet")),
+            None
+        )
```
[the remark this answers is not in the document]
BigRoy: Still pending code change?

client/ayon_maya/api/plugin.py
```diff
@@ -996,14 +995,9 @@ def remove(self, container):
...
-        namespace = container["namespace"]
-        associated_namespace_node = next((
-            associated_node for associated_node
-            in cmds.ls(f"*{namespace}", type="objectSet")),
-            None
-        )
-        if associated_namespace_node:
-            cmds.lockNode(associated_namespace_node, lock=False)
+        if "rig" in node.lower():
+            self._remove_linked_animation_instance(container)
```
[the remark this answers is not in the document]
BigRoy: Still pending code change?

> If we do it this way, this also means that we need to imprint the product type as container data

No, we'd query product type from the representation id.

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -239,12 +248,68 @@ def update(self, container, context):
...
+        )
+        product_type = None
+        if representation_id:
+            context: dict = get_representation_context(
+                project_name, representation_id
+            )
+            product_type: str = context["product"]["productType"]
+
+        if product_type == "rig":
+            # No special handling needed for non-rig containers
```
BigRoy, line 264:
```suggestion
            # Special handling needed for rig containers
```

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -239,12 +248,68 @@ def update(self, container, context):
...
+            self._remove_rig(container)
+            return
+
+        super().remove(container)
+
+    def _remove_rig(self, container):
```
BigRoy, line 270: Please add a docstring of what special behavior is needed for the rig products, so it's clear what this method is trying to do.

server/settings/loaders.py
```diff
@@ -110,6 +110,9 @@ class ReferenceLoaderModel(BaseSettingsModel):
     display_handle: bool = SettingsField(
         title="Display Handle On Load References"
     )
+    lock_animation_instance_on_load: bool = SettingsField(
+        title="Lock Animation Instance on Rig Load"
+    )
```
BigRoy, line 115: @LiborBatek how do you feel about this naming `lock_animation_instance_on_load`? Is that clear enough?

It's actually the _rig_ being loaded and the animation instance being created - so it feels a bit mixed.

client/ayon_maya/api/lib.py
BigRoy:
```suggestion
    has_locked = cmds.lockNode(node, query=True, lock=True)[0]
    cmds.lockNode(node, lock=False)
```

client/ayon_maya/api/lib.py
```diff
@@ -4247,7 +4246,7 @@ def _get_parent(reference_node):
     return parents
 
 
-def get_attr(node, attr, default=None):
+def get_instance_node_attr(node, attr, default=None):
```
BigRoy, line 4249: I understand you're trying to avoid the conflict with `get_attribute` but in reality this code has nothing to do with an `instance node`. @antirotor can you guide along perhaps - I'm too short on time to check what's a decent name and code structure here - and I feel with just a quick "heads up" from my end in the reviews it's just costing Kayla more time than needed.

PR #343, opened 2025-09-16
Update validate_scene_set_workspace.py

tokejepsen:
## Changelog Description
Fix extended paths on Windows for workspace validation.

When using long paths >260 characters, `os.path.realpath` will return the prefix `\\?\` due to being an extend path.

## Testing notes:
1. Make a long path to a workfile >260 characters.
2. Validate the workspace by publishing.

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_scene_set_workspace.py
+++ b/client/ayon_maya/plugins/publish/validate_scene_set_workspace.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 
 import maya.cmds as cmds
 from ayon_core.pipeline.publish import (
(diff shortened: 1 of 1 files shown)


moonyuet (commented):
Tested successfully in Maya with windows.
Haven't tested in other platforms and remote drives

tokejepsen:
> This likely works - but I hardly think that with too long paths on windows this is the ONLY place you'll be facing issues. With super long paths even something as simple as a file transfer to publish destination may fail (if that file exists in that path that is, like e.g. textures or maybe XGEN files?)

You are right, that is not the only place it's an issue, but due to being stuck in OpenPype and limited time, I can't transfer other fixes atm.

> So, in that scenario - if it does.. what's the error or issue you face? Does the splitdrive logic fail? or?

I think the problem was that `scene_name` was over path length limit and `root_dir` was under, so `scene_name` got extended prefix but `root_dir` did not.

BigRoy:
> I think the problem was that scene_name was over path length limit and root_dir was under, so scene_name got extended prefix but root_dir did not.

I see - I think the better fix here is using `pathlib.Path`

```python
Path(workfile).parent == Path(work_dir)
```
Making the check just this ☝️

PR #344, opened 2025-09-17
Host: Implement application information method

iLLiCiTiT:
## Changelog Description
Implement `get_app_information` on host class.

Currently the method is used with https://github.com/ynput/ayon-core/pull/1445 to be stored into workfile entity.

## Testing notes:
1. Validate the implementation is actually correct.

### DIFF
--- a/client/ayon_maya/api/pipeline.py
+++ b/client/ayon_maya/api/pipeline.py
@@ -80,6 +80,13 @@ class MayaHost(HostBase, IWorkfileHost, ILoadHost, IPublishHost):
         super(MayaHost, self).__init__()
         self._op_events = {}
 
+    def get_app_information(self):
+        from ayon_core.host import ApplicationInformation
+
+        version = cmds.about(version=True)
+
+        return ApplicationInformation("Maya", version)
+
     def install(self):
         project_name = get_current_project_name()
         project_settings = get_project_settings(project_name)


client/ayon_maya/api/pipeline.py
```diff
@@ -80,6 +80,13 @@ def __init__(self):
         super(MayaHost, self).__init__()
         self._op_events = {}
 
+    def get_app_information(self):
+        from ayon_core.host import ApplicationInformation
+
+        version = cmds.about(version=True)
+
+        return ApplicationInformation("Maya", version)
```
BigRoy, line 88: I see why this may be coming from the host like this - but admittedly I'd say that storing the AYON app name and tools is crucial too - regardless of what this information presents.

As soon as we start getting very accurate about which maya version is loaded it also becomes the time that suddenly I want to also store all the relevant plug-ins that are loaded or required for file and their exact versions, etc. Which I feel this method doesn't seem to really allow with it's return type? or?

Are we expecting a host name and a 'version' number all being relevant as 'app information'?

Other than that, the query itself seems fine.

iLLiCiTiT: > but admittedly I'd say that storing the AYON app name and tools is crucial too - regardless of what this information presents.

App name is stored [here](https://github.com/ynput/ayon-core/tree/enhancement/1443-allow-host-addons-to-store-dcc-version-in-workfile-metadata#diff-abb7c3fe03cddb36a29b179d98d01b031468c7d6503deaf5a693fc1b7f5bdb61R1566)?

Tools are not in current requirements.

BigRoy: > Tools are not in current requirements, but could be added in future too.

I see. Adding Tools will be very useful to an env var - by doing so, whenever you switch context we can also start recognizing if the current session has the correct tools applied that would also match the new context you're moving into - and if not, log a warning to the user. As per: https://github.com/ynput/ayon-core/issues/826

PR #360, opened 2025-10-20
Chore: Disable workspace validator on farm

kalisp:
## Changelog Description
Validator checks if current workfile is inside of `work` area which makes only sense locally, not on a farm.
Without this change remote publishing of `pointcache` would fail as in it Publish process runs in Maya host (headlessly) which triggered this plugin.

## Testing notes:
1. best would be test it with https://github.com/ynput/ayon-royalrender/pull/9 (without this Maya PR remote publish would fail)


moonyuet (approved):
Looks good to me. And it is still there when we are publishing locally.

PR #366, opened 2025-11-19
Make extract skeletal mesh alembic optional

### LINKED ISSUE #365: YN-0201: Maya: Extract Unreal Skeletal Mesh

 

there happens to be more some confusion than bug atm. Users struggle to put correct entities into `rig` subsets like `skeletalMesh` etc.

  

Example maya scene was provided so the client evaluate their issues and will report back if persisting...

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1437496187455279246)

_This issue was automatically created from AYON ticket [YN-0201](https://support.ayon.app/projects/Active_Clients/overview)_

moonyuet:
## Changelog Description
This PR is to make extract skeletal mesh alembic as an optional extractor
Resolve https://github.com/ynput/ayon-maya/issues/365

This PR is draft as it might deal with a list of things needed to be fixed in skeletal mesh workflow from maya to unreal.
And the title would be probably renamed
Build and install the addon with this branch before testing.

## Testing notes:
1. Launch Maya
2. Tweak the setting in `enhancement/YN-0201_Maya_Extract-Unreal-Skeletal-Mesh`
3. Create Skeletal Mesh Instance
4. You can find the Extract Skeletal Mesh Alembic can be turned on/off
5. Publish


LiborBatek (commented):
I have tried two times and built the addon package too but no luck and I cant see any `Extract Alembic` knob visible in the publisher UI..

Should I do something else to see it?

image

moonyuet:
@LiborBatek can you build and install addon with this branch and test it again?

LiborBatek (changes requested):
Even tho the settings is present in the addons setting...

Image

I cant find it still while been in the Publisher UI and having the instance selected (also recreated all from scratch didnt work either)

Image

...maybe the settings leads to a wrong instance type?? e.g. rig instead? (as it has also attributes regarding the skeletonMesh - just my wild guess without any evidence :)

moonyuet:
> Even tho the settings is present in the addons setting...
> 
> Image
> I cant find it still while been in the Publisher UI and having the instance selected (also recreated all from scratch didnt work either)
> 
> Image
> ...maybe the settings leads to a wrong instance type?? e.g. rig instead? (as it has also attributes regarding the skeletonMesh - just my wild guess without any evidence :)


It's not about the setting and the extractor does not show up due to the inheritance of the wrong class which is not for setting the extractor to be optional. Fixed in [b064c71](https://github.com/ynput/ayon-maya/pull/366/commits/b064c71defceeedc1ab38ab0005f40fa9abf8050)

LiborBatek (approved):
Now I can produce and also see the optional `ABC` export/repre

Image


Just a small note while inspecting the resulting product I was not able to produce an asset with geometry (I have tried the top GRP first but also using exact geometries in the `geometry` set didnt help) ...but I guess its not related to this PR??

here you can see empty groups instead of actual geometries (while inspecting the Outliner - highlighted nodes)
Image

moonyuet:
> Just a small note while inspecting the resulting product I was not able to produce an asset with geometry (I have tried the top GRP first but also using exact geometries in the `geometry` set didnt help) ...but I guess its not related to this PR??

It is not related to this PR but definitely we need another issue for the fix.

PR #372, opened 2025-12-02
🏛️Product base types: add support to creators

antirotor:
## Changelog Description
This is adding product base types support to existing creators. In this stage, it only mimics the product types. This has to be done in all DCC before further changes can be done in ynput/ayon-core to finalize the support.

Additional PRs to Maya should check the usage of `product_type` as family - plugins shouldn't rely on having product type as family because if/when this is customized, workflow could stop work or behave unpredicable.

## Testing notes:
Everything should work as before. With https://github.com/ynput/ayon-core/pull/1306 you should not see creator related warning when publisher opens.

### DIFF
--- a/client/ayon_maya/api/plugin.py
+++ b/client/ayon_maya/api/plugin.py
@@ -323,7 +323,8 @@ class MayaCreator(Creator, MayaCreatorBase):
                 product_type=self.product_type,
                 product_name=product_name,
                 data=instance_data,
-                creator=self,)
+                creator=self,
+            )
             self._add_instance_to_context(instance)
 
             self.imprint_instance_node(instance_node,


client/ayon_maya/plugins/create/create_camera.py
```diff
@@ -11,8 +11,10 @@ class CreateCamera(plugin.MayaCreator):
     identifier = "io.openpype.creators.maya.camera"
     label = "Camera"
     product_type = "camera"
+    product_base_type = "camera"
     icon = "video-camera"
 
+
```
iLLiCiTiT:
```suggestion
```

client/ayon_maya/plugins/create/create_workfile.py
```diff
@@ -67,11 +72,12 @@ def create(self):
...
-                folder_entity,
-                task_entity,
-                variant,
-                host_name,
+                project_name=project_name,
+                folder_entity=folder_entity,
+                task_entity=task_entity,
+                variant=variant,
+                host_name=host_name,
+                product_base_type=self.product_base_type
```
iLLiCiTiT:
```suggestion
```

iLLiCiTiT (approved):

PR #373, opened 2025-12-03
Feature/eng 3362/fix color management prefs disable

rhdutta:
## Changelog Description

Add support for `activate_host_color_management` setting in Maya's `set_colorspace()` function. When this setting is `false`, Maya's color management is now explicitly disabled. 

Previously, only the `workfile.enabled` setting was checked, which would simply skip color management configuration but not disable it, leaving Maya with its default (enabled) state. This allows studios to explicitly disable color management via AYON settings rather than relying on Maya's default preferences.

## Testing notes:

1. Set Maya imageio settings: `activate_host_color_management: false`, `workfile.enabled: false`
2. Launch Maya through AYON - verify color management is disabled in Preferences > Color Management
3. Set Maya imageio settings: `activate_host_color_management: true`, `workfile.enabled: false`
4. Launch Maya through AYON - verify color management state is unchanged (Maya defaults)
5. Set Maya imageio settings: `activate_host_color_management: true`, `workfile.enabled: true`
6. Launch Maya through AYON - verify color management is enabled with the configured render space, display, and view settings

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -3496,6 +3496,12 @@ def set_colorspace(project_settings=None):
         cmds.colorManagementPrefs(edit=True, cmEnabled=False)
         return
 
+    if not imageio["workfile"]["enabled"]:
+        log.info(
+            "AYON Maya Color Management settings for workfile are disabled."
+        )
+        return
+        
     # set color spaces for rendering space and view transforms
     def _colormanage(**kwargs):
         """Wrapper around `cmds.colorManagementPrefs`.


BigRoy (commented):
I see the reasoning - but I'm wondering if it was indeed the intend to "fully disable it".

Usually disabling OCIO management from AYON meant that it's 'unmanaged' by AYON and not that it was force disabled. @jakubjezek001 @antirotor thoughts? I'm fine with having a way to force disable OCIO in Maya - I just wonder what the right 'settings' would be like?

E.g. globally having host management **enabled** but workfile flag disabled I think makes more sense to be the thing that sounds like "manage it, but disable it explicitly". But now this PR makes it seems like if host management is **disabled** then force disable it always?

PR #382, opened 2025-12-08
Load Arnold standin: static ass file (single frame) should not be detected as sequence

### LINKED ISSUE #381: YN-0297: Maya: Load arnold standin incorrectly thinks it is a sequence

Maya: Load arnold standin incorrectly thinks it is a sequence.

  

it seems that

https://github.com/ynput/ayon-maya/blob/d3770555bea9fdc6c89ebe283d5c9de2fba615d3/client/ayon_maya/plugins/load/load_arnold_standin.py#L19

  

returns `True` when receiving:

`["assMain_v006.ass"]`,

maybe thinking that `v006` is from a collection?

`[<Collection "assMain_v%03d.ass [6]">]` 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1447601560795480124)

_This issue was automatically created from AYON ticket [YN-0297]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=4b125380d48011f0ba4597bbe1a3c31f)

BigRoy:
## Changelog Description

Load Arnold standin: static ass file (single frame) should not be detected as sequence

A published file like `assMain_v006.ass` should not be detected as sequence and hence should be 'held' when loaded.

Fix #381 

## Testing notes:

1. Publish static ass file.
2. It should be loaded as single frame (not time-dependent sequence)

### DIFF
--- a/client/ayon_maya/api/plugin.py
+++ b/client/ayon_maya/api/plugin.py
@@ -1,5 +1,8 @@
+"""Plugin for Ayon Maya API."""
+from __future__ import annotations
 import json
 import os
+from typing import Any, Optional
 
 import ayon_api
 import qargparse
@@ -314,11 +317,14 @@ class MayaCreator(Creator, MayaCreatorBase):
         with lib.undo_chunk():
             instance_node = cmds.sets(members, name=product_name)
             instance_data["instance_node"] = instance_node
+
+            # product base type support is added in ayon-core 1.7.0
             instance = CreatedInstance(
-                self.product_type,
-                product_name,
-                instance_data,
-                self)
+                product_type=self.product_type,
+                product_name=product_name,
+                data=instance_data,
+                creator=self,
+            )
             self._add_instance_to_context(instance)
 
             self.imprint_instance_node(instance_node,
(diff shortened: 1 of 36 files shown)


client/ayon_maya/plugins/load/load_arnold_standin.py
```diff
@@ -14,12 +15,11 @@
 from ayon_maya.api import plugin
 
 
-def is_sequence(files):
-    sequence = False
-    collections, remainder = clique.assemble(files, minimum_items=1)
+def is_sequence(files: list[str]) -> bool:
+    collections, _remainder = clique.assemble(files)
```
moonyuet:
```suggestion
    collections, _remainder = clique.assemble(files, patterns=[clique.PATTERNS["frames"]])
```
do we need to check on the frames pattern or?

BigRoy: I've tweaked the logic - now it only cares about files of same extension - and if it happens to find multiple frames of that on disk AND it matches in any form to be a collection according to `clique` then we'll assume it's a sequence. That way, we don't accidentally say it's a sequence if there happen to be other files next to it that are....

Anyway, the frames pattern I think is redundantly specific in this scenario - so rather not lock it down further until we face an issue.

moonyuet:
@LiborBatek can you test it?

LiborBatek (approved):
As far as I can tell it works as expected... did full publish from scratch and loading...

image

Issue #375, opened 2025-12-04
YN-0286: Maya render add Render Targets support


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Add the Render targets in Maya for Local Render, Farm Render, Use existing frames, Use existing frames (farm) like is supported in e.g. Nuke and Houdini. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1446169716245794857)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0286]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=9f4dfd20d12c11f0a3b87d966928fe7a)

PR #383, opened 2025-12-11
Render: Add render targets like "Local Render" and "Use Existing Frames"

BigRoy:
## Changelog Description

Render: Add render targets like "Local Render" and "Use Existing Frames".

Fix https://github.com/ynput/ayon-maya/issues/375

This has some overlap with https://github.com/ynput/ayon-blender/pull/217 in Blender because I've tried to set up the local render instances in a 'similar' manner using the data from the `expectedFiles` like the farm jobs would. This way, we can hopefully ensure as much as possible that both flow through the same publishing steps. The end goal would be to unify in such a way that potentially - it could even be shared or inherited from to avoid needing to maintain unique code flows for this between Maya, Blender, etc. and likely we should look into then also matching similar in Fusion, Nuke, Houdini, etc.

## Testing notes:

Requires: https://github.com/ynput/ayon-deadline/pull/223

1. [ ] The different Render Targets "Local Render", "Use Existing Frames" and "Farm Render" should work:
    - [x] Local Render
    - [ ] Use Existing Frames
    - [ ] Farm Render
3. [ ] Should work correctly with multiple renderlayers (it should render the correct layer for each)

The code changes may also affect review playblasts, so also test:
- [ ] "Reviews" (playblasts) still work as intended, with correct publishes and reviews.

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -1850,7 +1850,8 @@ def get_container_members(container, include_reference_associated_nodes=False):
     This includes the nodes from any loaded references in the container.
 
     Arguments:
-        container (str | dict): container data or name of container node
+        container (str | dict): The container name or a container dictionary
+            with the "objectName" key.
         include_reference_associated_nodes (bool): whether to include the
             associated nodes of references, like those produced from
             referencing with `groupReference`. This is disabled by default,
@@ -2252,12 +2253,13 @@ def get_container_transforms(container, members=None, root=False):
     transform is stored in the container information
 
     Args:
-        container (dict): the container
+        container (dict | str): the container
         members (list): optional and convenience argument
         root (bool): return highest node in hierarchy if True
 
     Returns:
-        root (list / str):
+        list[str] | str: List of highest nodes, or first entry if
+            root argument is True
     """
 
     if not members:
@@ -2377,7 +2379,15 @@ def remove_other_uv_sets(mesh):
 
 
 def get_node_parent(node):
-    """Return full path name for parent of node"""
+    """Return full path name for parent of node
+
+    Arguments:
+        node (str): The node path of the node to get the parent for.
+
+    Returns:
+        str | None: The full path name of the parent node or None.
+
+    """
     parents = cmds.listRelatives(node, parent=True, fullPath=True)
     return parents[0] if parents else None
 
--- a/client/ayon_maya/plugins/workfile_build/load_placeholder.py
+++ b/client/ayon_maya/plugins/workfile_build/load_placeholder.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from maya import cmds
 
 from ayon_core.pipeline.workfile.workfile_template_builder import (
@@ -6,6 +7,7 @@ from ayon_core.pipeline.workfile.workfile_template_builder import (
 )
 from ayon_maya.api.lib import (
     get_container_transforms,
+    get_highest_in_hierarchy,
     get_node_parent,
     get_node_index_under_parent
 )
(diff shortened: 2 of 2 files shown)


LiborBatek:
Once triggering the local render of my render instance Im getting this error (but there is shown frame buffer poping up for each frame of the sequence but empty and following traceback in maya console:
```
// Traceback (most recent call last):
//   File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2510271030_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
//     runner(*args)
//   File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.5.1+dev\ayon_maya\plugins\publish\extract_render.py", line 76, in process
//     with contextlib.ExitStack() as stack:
//   File "C:\Program Files\Autodesk\Maya2026\Python\Lib\contextlib.py", line 601, in __exit__
//     raise exc_details[1]
//   File "C:\Program Files\Autodesk\Maya2026\Python\Lib\contextlib.py", line 586, in __exit__
//     if cb(*exc_details):
//        ^^^^^^^^^^^^^^^^
//   File "C:\Program Files\Autodesk\Maya2026\Python\Lib\contextlib.py", line 144, in __exit__
//     next(self.gen)
//   File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.5.1+dev\ayon_maya\plugins\publish\extract_render.py", line 28, in hidden_render_view
//     cmds.deleteUI("renderViewWindow", window=True)
// RuntimeError: Object 'renderViewWindow' not found.

```

BigRoy:
@LiborBatek can you try again now? I wonder why it didn't initialize that UI for you with that command though. 🤔 Anyway, it should pass that now.

BigRoy:
@LiborBatek I've made some changes to support the review.

It may affect a few things. So, can you make sure to test:
- [ ] Farm render still work as intended
- [ ] Local renders now work as intended with reviews
- [ ] "Reviews" (playblasts) still work as intended, with correct publishes and reviews.

client/ayon_maya/plugins/publish/extract_render.py
```diff
@@ -0,0 +1,130 @@
...
+        # renderers will render into a `tmp` subfolder under the 'images' file
+        # rule directory. We need to move the files up a folder.
+        image_directory: str = os.path.join(
+            cmds.workspace(query=True, rootDirectory=True),
+            cmds.workspace(fileRuleEntry="images"),
+        )
+        expected_files: list[dict[str, list[str]]] = (
+            instance.data["expectedFiles"]
+        )
+        for _aov, filepaths in expected_files[0].items():
```
iLLiCiTiT:
```suggestion
        for filepaths in expected_files[0].values():
```

LiborBatek (approved):
In essence now both `Farm render` and `Local Render` been working and also `Review` still working as expected too...

I was able to produce render products in both scenarios...

There is a some folder naming inconsistency when speaking about staging aka work render folders as seen here

when using `Local render` its different than when in `Farm render` mode (pointed in green)
image

Not sure whether its an issue or not but it def. adds some unnecessary clutter into folders imho.

I was also trying to test out the `Local Use existing frames` feature but was not able to succeed as it seems it expect some existing rendered frames but it targets the following version folder which obviously do not exists yet... again not sure the intended workflow here but its not usable atm.

Here are my existing folders before trying to use such `render` publish
image

and when hitting the issue while trying to publish:

image



Still can be approved as it is working in essence... speaking of those main `Farm` and `Local` renders...

BigRoy:
> Not sure whether its an issue or not but it def. adds some unnecessary clutter into folders imho.

This is due to one rendering from the published workfile, and the other from the source workfile and both rendering with the `<Scene>` token, so it is to be expected.

> I was also trying to test out the Local Use existing frames feature but was not able to succeed as it seems it expect some existing rendered frames but it targets the following version folder which obviously do not exists yet... again not sure the intended workflow here but its not usable atm.

The intended workflow I understood from client was if they'd have triggered the Render Sequence or alike manually. Not seeing many users do, but that's what it's for. 🤷‍♂️ 

So the validation is again, correct.
I'll merge this since the feature set is there and we'll have to await client feedback to see if it matches their workflow enough after.

client/ayon_maya/plugins/publish/extract_render.py
BigRoy: Sorry - missed these comments. Already merged by accident. Follow up: https://github.com/ynput/ayon-maya/pull/391

PR #387, opened 2025-12-15
allow keeping maya's cachedPlayback Cache

vincentullmann:
## Changelog Description
Refactor the `_applied_camera_options` method to only apply and revert those attributes where the values differ from the current ones.
The generation of thumbnail would use the current frame instead of the start frame. 

This allows maya to keep the "cachedPlayback" Cache and thus can impove playblast performance

## Testing notes:
1. open a heavy maya scene (eg.: large image planes)
2. scrub trough the timeline to populate the playback cache
3. make sure your camera settings match the ExtractPlayblast settings!
4. publish a playblast

### DIFF
--- a/client/ayon_maya/vendor/python/capture.py
+++ b/client/ayon_maya/vendor/python/capture.py
@@ -652,37 +652,46 @@ def _independent_panel(width, height, off_screen=False):
 @contextlib.contextmanager
 def _applied_camera_options(options, panel):
     """Context manager for applying `options` to `camera`"""
+    if not options:
+        yield
+        return
 
     camera = cmds.modelPanel(panel, query=True, camera=True)
-    options = dict(CameraOptions, **(options or {}))
+    options = {**CameraOptions, **options}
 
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
I have performed few test rounds and didnt spotted any issues...also the thumbnails been captured from the current time/frame the user been at...cool! as seen on my thumbs below

image

Note: even trying to stress test it I couldnt make a clear speed up difference but looks ok to me...

vincentullmann:
most of this PR deal with changes to a vendor lib, so I don't think this PR can be merged as is.

I'm happy to reopen the thumbail-frame change as a new clean PR.
Coming back to this with a fresh view I recon it might be worth adding a bit of control, so clients can choose from a couple of options such as "current frame, first frame, last frame, mid frame, custom frame" in case they want to enforce consitent thumbnail frames.
I can see how some might not like the idea of getting different thumbnail frames on each published version depending on which frame the user was at the time of publish...

Issue #371, opened 2025-12-02
YN-0250: Maya Support publishing multiple yeti rigs from one workfile


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

When publishing multiple yetiRig grooms from one workfile it won't work because it requires an `input_SET` to exist in each yeti instance.

  

We should fix it so that it can e.g. be prefixed like `yetRigMain_input_SET` or alike so that each can be published fine together. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0250]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=14a08510cba211f09260f32967801451)

PR #389, opened 2026-01-13
Yeti: Support publishing multiple Instances in one workfile 

moonyuet:
## Changelog Description
This PR is to add support for publishing multiple yeti rig instances in one workfile
Resolve https://github.com/ynput/ayon-maya/issues/371#

n/a

## Testing notes:
1. Create multiple Yeti Rig instances
2. Publish


client/ayon_maya/plugins/publish/validate_yeti_rig_input_in_instance.py
```diff
@@ -21,26 +21,26 @@ class ValidateYetiRigInputShapesInInstance(plugin.MayaInstancePlugin,
...
-        input_nodes = cmds.ls(cmds.sets(input_set, query=True), long=True)
-        dag = cmds.ls(input_nodes, dag=True, long=True)
-        shapes = cmds.ls(dag, long=True, shapes=True, noIntermediate=True)
-
-        # Allow publish without input meshes.
-        if not shapes:
-            cls.log.debug("Found no input meshes for %s, skipping ..."
-                          % instance)
-            return []
+        for input_set in instance.data["yeti_sets"].values():
```
BigRoy: Expect dedicated `input_SET` key, do not iterate just "all sets". See other comments.

This should also invalidate if `yeti_sets.get("input_SET") is None` because it'd mean no set was found whatsoever.

client/ayon_maya/plugins/publish/collect_yeti_rig.py
```diff
@@ -52,48 +48,81 @@ def collect_input_connections(self, instance):
...
 
         return inputs
 
+    def get_yeti_sets(self, instance):
+        """Get all Yeti sets from the instance
+
+        Args:
+            instance (pyblish.Instance): instance to collect from
+        Returns:
+            dict: mapping of set names to their members
```
BigRoy:
```suggestion
            dict[str, str]: mapping of set names to their members
```

client/ayon_maya/plugins/publish/collect_yeti_rig.py
```diff
@@ -85,15 +86,42 @@ def collect_input_connections(self, instance):
...
+        """Get all Yeti sets from the instance
+
+        Args:
+            instance (pyblish.Instance): instance to collect from
+        Returns:
+            dict[str, str]: mapping of set names to their members
+        """
+        # Find required sets by suffix
+        searching = {"input_SET"}
+        found: dict[str, str] = {}
```
BigRoy: For this type hint, can you add `from __future__ import annotations` to the top of the file please?

client/ayon_maya/plugins/publish/validate_yeti_rig_input_in_instance.py
```diff
@@ -21,25 +21,26 @@ class ValidateYetiRigInputShapesInInstance(plugin.MayaInstancePlugin,
...
+        input_set = yeti_sets["input_SET"]
         input_nodes = cmds.ls(cmds.sets(input_set, query=True), long=True)
         dag = cmds.ls(input_nodes, dag=True, long=True)
         shapes = cmds.ls(dag, long=True, shapes=True, noIntermediate=True)
 
         # Allow publish without input meshes.
         if not shapes:
             cls.log.debug("Found no input meshes for %s, skipping ..."
-                          % instance)
+                        % instance)
```
BigRoy: Fix indent
```suggestion
                          % instance)
```

BigRoy:
I pushed some changes, sorry my OCD kicked in on the indentations being just off and Github suggestions made it really hard to get them right it seemed. So I quickly went into my IDE to apply the changes, and noticed another bug there with the `cmds.sets(..., query=True)` call having gone missing somewhere, so I pushed that fixed here too.

I think this covers everything that I can spot quickly. Sorry for pushing directly here.

LiborBatek (commented):
So now I was able to produce publishes for both `yetiRig`s at once (Main, Alter variants)...

Screenshot 2026-01-14 172159

Which is great but there is an issue that the `yetiRig` brings unneccessary items with it (connected entities from the rigged char which was used for connections of the groom mesh duplicates)

before this PR it looked like this:


Screenshot 2026-01-14 172745

while now it has all the rigged parts together with the yeti_setup:


Screenshot 2026-01-14 172841


I have also tested the older workfiles (before this PR) and it still works as it doesnt have any namespaces withing the INPUT_SET while my new workfile with two yeti setups and new naming produces this product containing pretty much any connected entities.

Not really sure why and if anyhow connected but occurs atm

LiborBatek:
@BigRoy yeah, here you have it...the former setup before publishing and second scrn when loaded (this case been correct)

Screenshot 2026-01-16 094438
and when loaded:

Screenshot 2026-01-16 094615


And the recent non working setup (using namespaces within the `input_SET` name..btw reverting to the old naming didnt make any change and still malfunctions)

Screenshot 2026-01-16 094837

and when such publish been loaded it brings also the source node connected rig asset (groom geo been connected via transforms and shape node's `outmesh` to `inmesh` as suggested in the user docs and been working until this PR)

Screenshot 2026-01-16 095942


Note: we had a live call debugging session with @moonyuet yesterday but didnt come up with any source of the issue (we tried to just use single yeti setup within a scene, renaming input sets but no luck so far...

client/ayon_maya/plugins/publish/extract_yeti_rig.py
```diff
@@ -166,8 +173,8 @@ def process(self, instance):
         nodes = instance.data["setMembers"]
         resources = instance.data.get("resources", {})
         with disconnect_plugs(settings, members):
-            with yetigraph_attribute_values(resources_dir, resources):
-                with lib.attribute_values(attr_value):
+            with (resources_dir, resources):
+                with lib.atyetigraph_attribute_valuestribute_values(attr_value):
```
BigRoy: Typo? :) Pretty sure this is broken now. :)

PR #397, opened 2026-02-17
Fix Extract Playblast not triggering when review is attached to another instance

BigRoy:
## Changelog Description

Fix Extract Playblast not triggering when review is attached to another product (e.g. `pointcache` instance object set is inside the `review` instance object set)

Mentioned [here](https://discord.com/channels/517362899170230292/563751989075378201/1473218121350381642)

## Testing notes:

1. Playblast should trigger correctly in all scenarios.

### DIFF
--- a/client/ayon_maya/plugins/publish/collect_review.py
+++ b/client/ayon_maya/plugins/publish/collect_review.py
@@ -70,12 +70,13 @@ class CollectReview(plugin.MayaInstancePlugin):
             data = reviewable_inst.data
 
             self.log.debug(
-                'Adding review family to {}'.format(reviewable_product)
+                'Adding review families to {}'.format(reviewable_product)
             )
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
It does produce `Review` in all circumstanes but its worth mentioning that when using the legacy way of attaching the review to the particular product ( `model` in my case) it wont attach it at all but produce standalone `Review` version as like there is no any attachment made.

So reviewables are generated as supposed to but legacy attachement not working.

Also enclosing my scene outliner and publish report JSON.

Image

[maya-review-publish-report-260217-15-55.json.txt](https://github.com/user-attachments/files/25367832/maya-review-publish-report-260217-15-55.json.txt)

Note:
So this was my user fail setting it up wrong...it has been checked again and been working while correctly setup.

Issue #401, opened 2026-02-25
YN-0523: Maya: Referenced camera with attached image plane fails to extract camera


ynbot:
### Please describe the issue you have encountered?

Whenever you have a referenced camera that comes with an image plane attached underneath it in the reference than the Extract Camera for the camera instance will fail with error:

  

    
    # Error: RuntimeError: Unable to edit the image plane.
    

  

Which is due to the [transfer image plane logic here](https://github.com/ynput/ayon-maya/blob/ccdfc1ca2cbbb3e4f38f8034add02c4f131f66f5/client/ayon_maya/plugins/publish/extract_camera_mayaScene.py#L281-L294) where the attach image plane logic disallows to re-attach to the referenced camera after. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0523]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=060a55a0124511f19287c3def1f491ba)

PR #404, opened 2026-02-27
Camera: Check if image plane is referenced during detaching in the context

moonyuet:
## Changelog Description
This PR is to check if image plane is referenced during detaching in the context. If it is the reference node, it would skip the detaching process. 
Resolve https://github.com/ynput/ayon-maya/issues/401

I am not sure if it is right approach to do so make it draft before final decision is made.

## Testing notes:
1. Create Camera with image plane
2. Publish Camera
3. Loaded back the published camera
4. Then publish 'camera' product type with that loaded camera.


BigRoy (approved):
Awesome.

PR #405, opened 2026-03-03
Render: Product type in product name

iLLiCiTiT:
## Changelog Description
Fir product names in render create plugin.

Render create plugin makes sure that `"render"` is used as product base type and product type for product name value.

For product type, the create plugin makes sure that `get_product_type_items` returns one item with `render` product type -> make sure product base type is not passed by publisher.

Override `get_product_name` to pass `"render"` as product base type.

## Testing notes:
1. Render instances don't use `renderlayer` as product type in product name.

### DIFF
--- a/client/ayon_maya/api/plugin.py
+++ b/client/ayon_maya/api/plugin.py
@@ -2,7 +2,6 @@
 from __future__ import annotations
 import json
 import os
-from typing import Optional
 
 import ayon_api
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Thanks @iLLiCiTiT - this seems to work, but needs to likely be moved to `RenderlayerCreator`.

PR #424, opened 2026-04-27
Improve debug message for validate frame range

### LINKED ISSUE #423: YN-0586: Adding Debug message to FPS,FrameRange Validators for lower confusion

If you are in a task, it should validate against the task admittedly. So there shouldn't really be any confusion - but it may be worth adding a debug log message here and there where it may take either of the two so it's clear which one it took.,

  

e.g from Folder level  or Task level.

  

from the client:

  

Frame range validator error message is now confusing user by offering incorrect info (there is mentioned term 'folder' even tho the 'task' level been used. 

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0586]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=93f9c3c0220b11f1a9723959ba84feae)

moonyuet:
## Changelog Description
This PR is to improve debug message for validate frame range so that the users won't be confused where they should correct their frame range settings (Should they correct through task or folder). If there is no task specific, the frame range should be correct through the folder entity.

Resolve: https://github.com/ynput/ayon-maya/issues/423

n/a

## Testing notes:
1. Launch Maya
2. Make sure you have frame range which does not align with the folder/task entity set in your project
3. Publish
4. Validation block
5. The message would show incorrect frame range at which task and folder path.

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_frame_range.py
+++ b/client/ayon_maya/plugins/publish/validate_frame_range.py
@@ -82,10 +82,10 @@ class ValidateFrameRange(plugin.MayaInstancePlugin,
             return
 
         checks = {
+            "frame start from context": (frame_start_handle, inst_start),
+            "frame end from context": (frame_end_handle, inst_end),
             "frame start": (frame_start, inst_frame_start),
             "frame end": (frame_end, inst_frame_end),
-            "frame start handle": (frame_start_handle, inst_start),
-            "frame end handle": (frame_end_handle, inst_end),
             "handle start": (handle_start, inst_handle_start),
             "handle end": (handle_end, inst_handle_end)
         }
--- a/client/ayon_maya/plugins/publish/validate_frame_range.py
+++ b/client/ayon_maya/plugins/publish/validate_frame_range.py
@@ -82,19 +82,17 @@ class ValidateFrameRange(plugin.MayaInstancePlugin,
             return
 
         checks = {
-            "frame start from context": (frame_start_handle, inst_start),
-            "frame end from context": (frame_end_handle, inst_end),
+            "frame start including handles": (frame_start_handle, inst_start),
+            "frame end including handles": (frame_end_handle, inst_end),
             "frame start": (frame_start, inst_frame_start),
             "frame end": (frame_end, inst_frame_end),
             "handle start": (handle_start, inst_handle_start),
             "handle end": (handle_end, inst_handle_end)
         }
-        for label, values in checks.items():
-            if values[0] != values[1]:
+        for label, (required, current) in checks.items():
+            if current != required:
                 errors.append(
-                    "{} on instance is {} but should be {}. ".format(
-                        label.title(), values[0], values[1]
-                    )
+                    f"{label.title()} on instance is {current} but should be {required}. "
                 )
 
         if errors:
@@ -102,7 +100,7 @@ class ValidateFrameRange(plugin.MayaInstancePlugin,
             folder_path = context.data.get("folderPath")
             if task_name:
                 report = (
-                    "Frame range settings do not match \n\n"
+                    "Frame range settings do not match"
                     f"task '{task_name}' in folder '{folder_path}'.\n\n"
                 )
             else:
--- a/client/ayon_maya/plugins/publish/validate_frame_range.py
+++ b/client/ayon_maya/plugins/publish/validate_frame_range.py
@@ -92,12 +92,24 @@ class ValidateFrameRange(plugin.MayaInstancePlugin,
         for label, values in checks.items():
             if values[0] != values[1]:
                 errors.append(
-                    "{} on instance ({}) does not match with the folder "
-                    "({}).".format(label.title(), values[1], values[0])
+                    "{} on instance is {} but should be {}. ".format(
+                        label.title(), values[0], values[1]
+                    )
                 )
 
         if errors:
-            report = "Frame range settings are incorrect.\n\n"
+            task_name = context.data.get("task")
+            folder_path = context.data.get("folderPath")
+            if task_name:
+                report = (
+                    "Frame range settings do not match \n\n"
+                    f"task '{task_name}' in folder '{folder_path}'.\n\n"
+                )
+            else:
+                report = (
+                    "Frame range settings do not match \n\n"
+                    f"folder '{folder_path}'.\n\n"
+                )
             for error in errors:
                 report += "- {}\n\n".format(error)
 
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
The validator correctly kicks in when either `folder` or `task` frame start/end attribute differs but the info message is unclear and the values listed are very convoluted and barely readible:

Image

Also besides the values issues above, its not very clear to the user on which entity level the values do not match (if on folder level or task level) which is the original reasoning behind this PR...to correctly refer in a clear way which values differ.

Image

Or maybe we should not care about the source in the info message at all and just inform that values differing from the project set ones... and just tweak the way the values are displayed (the above issue).

LiborBatek (commented):
In essence it works and informs correctly when relating to the values...there is just still some bit missing...and thats to distinguish among `task` and `folder` attribute...I mean it would be lovely to have it included and differentiate among those values...so user could get better idea...Im not saying its a mandatory but there was a such confusion (whether the correct values originating from the folder or task level...

here is my info msg mentioning `folder` term even tho it should be rather `task` as those values been taken from the task level...

Image

and the actual attribs and values in the project : (again note the folder vs task attrib) it would be handy to be able to distinguish among them properly and in the wording within the info msg:

Image

Otherwise the logic works correctly as I have tested multiple scenarios using handle values etc.

BigRoy:
It may be better to phrase the messaging as something like:

> Frame range settings do no match folder: '{folder_path}'.

OR

> Frame range settings do not match task '{task_name}' in folder '{folder_path}'.
> 
> - Frame start is 999 but should be 1001.
> - Frame end is 1010 but should be 1015.
> - FPS is 24 but should be 25.

Instead of duplicating the full task vs folder labeling per mismatch.
@LiborBatek thoughts?

moonyuet:
> It may be better to phrase the messaging as something like:
> 
> > Frame range settings do no match folder: '{folder_path}'.
> 
> OR
> 
> > Frame range settings do not match task '{task_name}' in folder '{folder_path}'.
> > 
> > * Frame start is 999 but should be 1001.
> > * Frame end is 1010 but should be 1015.
> > * FPS is 24 but should be 25.
> 
> Instead of duplicating the full task vs folder labeling per mismatch. @LiborBatek thoughts?

I personally prefer the latter one.

BigRoy:
> I personally prefer the latter one.

The OR is intended for the case where it may have a task set or not ;) so both would apply as the first line, depending on whether there's a task associated with the instance.

client/ayon_maya/plugins/publish/validate_frame_range.py
```diff
@@ -108,12 +92,24 @@ def process(self, instance):
...
+            if task_name:
+                report = (
+                    "Frame range settings do not match \n\n"
+                    f"task '{task_name}' in folder '{folder_path}'.\n\n"
+                )
+            else:
+                report = (
+                    "Frame range settings do not match \n\n"
+                    f"folder '{folder_path}'.\n\n"
+                )
```
BigRoy, line 110:
```suggestion
            if task_name:
                report = (
                    "Frame range settings do not match"
                    f" task '{task_name}' in folder '{folder_path}'.\n\n"
                )
            else:
                report = (
                    "Frame range settings do not match"
                    f" folder '{folder_path}'.\n\n"
                )
```

client/ayon_maya/plugins/publish/validate_frame_range.py
```diff
@@ -108,12 +92,24 @@ def process(self, instance):
         for label, values in checks.items():
             if values[0] != values[1]:
                 errors.append(
-                    "{} on instance ({}) does not match with the folder "
-                    "({}).".format(label.title(), values[1], values[0])
+                    "{} on instance is {} but should be {}. ".format(
+                        label.title(), values[0], values[1]
+                    )
```
BigRoy: These values being swapped even though the order of the values being the same in the phrase itself seems wrong?
```suggestion
                    "{} on instance is {} but should be {}. ".format(
                        label.title(), values[1], values[0]
                    )
```

Another approach would be to write the for loop as `for label, (required, current) in checks.items():` so that the phrase here just reads:
```
f"{label} on instance is {current} but should be {required}"
```

Which may make the code more readable too?

client/ayon_maya/plugins/publish/validate_frame_range.py
```diff
@@ -81,25 +81,9 @@ def process(self, instance):
...
-                          "match the one set on folder [ {} ]: "
-                          "{}/{}/{}/{} (handle/start/end/handle)".format(
-                              inst_end,
-                              frame_end_handle,
-                              handle_start, frame_start, frame_end, handle_end
-                          ))
-
         checks = {
+            "frame start from context": (frame_start_handle, inst_start),
+            "frame end from context": (frame_end_handle, inst_end),
```
BigRoy: I think this label is quite confusing - right? Shouldn't this be "Frame start including handles"?

Also, shouldn't _technically_ checking for just handle start and frame start tackle this same check? 🤔 Isn't this actually redundant to check additionally?

BigRoy (approved):
Code looks good to me - @LiborBatek can you test a few publishes? Can you test both e.g. pointcache products and renders as well?

LiborBatek:
I will give it a go...

...speaking of that info msg format, the shorter the better! totally agreed

LiborBatek (approved):
I have tested with multiple publish instance types (anim, pointcache, camera, render) and all seems fine.

LGTM as it is atm

Image

client/ayon_maya/plugins/publish/validate_frame_range.py
```diff
@@ -81,39 +81,33 @@ def process(self, instance):
...
-                          "match the one set on folder [ {} ]: "
-                          "{}/{}/{}/{} (handle/start/end/handle)".format(
-                              inst_end,
-                              frame_end_handle,
-                              handle_start, frame_start, frame_end, handle_end
-                          ))
-
         checks = {
+            "frame start including handles": (frame_start_handle, inst_start),
+            "frame end including handles": (frame_end_handle, inst_end),
```
BigRoy, line 86: I actually think we should exclude these. It just looks more confusing in the final report looking at @LiborBatek 's screenshot. Currently whenever start frame is wrong or handle start is wrong it will ALWAYS also display this.

Essentially, I don't see a case where this would be invalid if both frame and handle ranges are correct. As such, checking (and logging this) separately should be entirely redundant and is just outright confusing.

I noticed this was quickly merged after the approval by Libor. @moonyuet can you make a follow up PR that just removes this please?

PR #425, opened 2026-04-29
Ensure the scene without trailing collection after camera publish

### LINKED ISSUE #44: Bug: Maya - Publishing camera from renderlayer creates untitled collections

### Current Behavior:

When publishing a camera instance whilst currently in a render setup layer **and** having the default setting enabled for **Enable untitled collections when new objects are created** the baked camera (only used for the export) gets added into an untitled collection.

image

image

The maya setting is [documented here](https://help.autodesk.com/view/MAYAUL/2023/ENU/?guid=GUID-A622E999-38E6-46B1-939F-F8FA6D3D61BA).

### Expected Behavior:

1. Leave the scene unaltered. As such, make sure to leave no trailing collection.
2. Another question of course would be whether we'd even want to export from an active renderlayer to begin with - but I suppose that's a separate issue.

---

### Context manager

We could use a context manager to ensure the setting is disabled.
Either setting the option var, or the environment variable.

_Not preferred:_

```python
from maya import cmds
import contextlib


@contextlib.contextmanager
def rendersetup_use_untitled_collections_context(value):
    key = "renderSetup_useUntitledCollections"
    if cmds.optionVar(exists=key):
        original_value = cmds.optionVar(query=key)
    else:
        original_value = None
        
    try:
        cmds.optionVar(intValue=(key, int(value)))
        yield
    finally:
        # Revert original value
        if original_value is None:
            cmds.optionVar(remove=True)
        else:
            cmds.optionVar(intValue=(key, original_value))
```

Since **The environment variable takes precedence over the preference** I'd say we should use the environment variable instead.

_Preferred:_

```python
import os
import contextlib


@contextlib.contextmanager
def rendersetup_use_untitled_collections_context(value):
    key = "MAYA_RENDER_SETUP_USE_UNTITLED_COLLECTIONS"
    original_value = os.getenv(key)
        
    try:
        # Allow only 1 or 0 as string value
        os.environ[key] = str(int(bool(value)))
        yield
    finally:
        # Revert original value
        if original_value is None:
            del os.environ[key]
        else:
            os.environ[key] = value


# Example
with rendersetup_use_untitled_collections_context(False):
    cmds.polyCube()
```

### Version

3.18.2-nightly.4

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Create camera
2. Create camera instance
3. Create and activate a render setup layer
4. Have **Enable untitled collections when new objects are created** setting enabled
5. Publish camera

There will now be a trailing (invalid) collection generated from the publish.

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-7982](https://app.clickup.com/t/86bx4bjb0)]

moonyuet:
## Changelog Description
This PR is to ensure the scene does not have any trailing collection with the existing rendersetup after the camera publish with baked options enabled.
Resolve https://github.com/ynput/ayon-maya/issues/44
n/a

## Testing notes:
1. Create camera
2. Create camera instance
3. Create and activate a render setup layer
4. Have Enable untitled collections when new objects are created setting enabled
5. Publish camera

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_camera_mayaScene.py
+++ b/client/ayon_maya/plugins/publish/extract_camera_mayaScene.py
@@ -160,7 +160,7 @@ class ExtractCameraMayaScene(plugin.MayaExtractorPlugin,
             stack.enter_context(lib.maintained_selection())
             stack.enter_context(lib.evaluation("off"))
             stack.enter_context(lib.suspended_refresh())
-            stack.enter_context(rendersetup_use_untitled_collection())
+            stack.enter_context(rendersetup_disable_untitled_collection())
             if bake_to_worldspace:
                 baked = lib.bake_to_world_space(
                     transforms,
@@ -306,7 +306,7 @@ def transfer_image_planes(source_cameras, target_cameras,
 
 
 @contextlib.contextmanager
-def rendersetup_use_untitled_collection():
+def rendersetup_disable_untitled_collection():
     """Disable MAYA_RENDER_SETUP_USE_UNTITLED_COLLECTIONS environment
     variable during context to ensure no trailing collection after the publish.
     """
--- a/client/ayon_maya/plugins/publish/extract_camera_mayaScene.py
+++ b/client/ayon_maya/plugins/publish/extract_camera_mayaScene.py
@@ -160,6 +160,7 @@ class ExtractCameraMayaScene(plugin.MayaExtractorPlugin,
             stack.enter_context(lib.maintained_selection())
             stack.enter_context(lib.evaluation("off"))
             stack.enter_context(lib.suspended_refresh())
+            stack.enter_context(rendersetup_use_untitled_collection())
             if bake_to_worldspace:
                 baked = lib.bake_to_world_space(
                     transforms,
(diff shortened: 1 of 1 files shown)


client/ayon_maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -304,6 +305,23 @@ def transfer_image_planes(source_cameras, target_cameras,
                 _attach_image_plane(camera, image_plane)
 
 
+@contextlib.contextmanager
+def rendersetup_use_untitled_collection():
```
BigRoy: Function name sounds like the opposite of what it does.
```suggestion
def rendersetup_disable_untitled_collection():
```

BigRoy (approved):
Changes look good to me - @LiborBatek can you test?

client/ayon_maya/plugins/publish/extract_camera_mayaScene.py
moonyuet: Fixed in [f336d0d](https://github.com/ynput/ayon-maya/pull/425/commits/f336d0dc757b75484bbe1811e7ed31e41745fdfa)

LiborBatek (approved):
I have followed the testing steps and all been working as expected.
I was able to publish the `camera` and also load it into an empty scene without any traces of unwanted data...

LGTM

PR #427, opened 2026-04-30
Implementation of validator for scene unknown nodes

### LINKED ISSUE #228: Improve Publish Error Report when 'maya-unknown-nodes-save-error' happens due to unloaded addon like Arnold render

When trying to publish e.g. ```model``` and the workfile having ```.mb``` format and having some unknown nodes present (e.g. when accidentally unloaded mtoa plugin)  and AYON trying to force prefered ```.ma``` as representation for workfile publish will fail and with not much clear message whats going on.

Image

when inspected in the detail pane:

```
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\addons\maya_0.4.3\ayon_maya\plugins\publish\extract_model.py", line 91, in process
    cmds.file(path,
RuntimeError: This file contains unknown data which is preventing Maya from changing the file format.
Check if unsupported plug-ins are used in the scene. Remove the unsupported data and resave.
See this article for more detail.
https://www.autodesk.com/maya-unknown-nodes-save-error

```

### Suggested implementation?

It should user inform in a clear way whats the possible cause of the issue and propose a solution to it.

e.g.

```It happens you may miss some maya addons loaded so AYON cant change workfile file format due to the unknown nodes present in your scene.```

```Please resolve unknown nodes via loading apropriate maya plugin or remove such nodes first so saving action can be performed.```

You get the idea.

Maybe its worth to think about possible logic to not change file format if changing could not be performed due to these circumstances. (not sure about it tho)

_No response_

### PR DESCRIPTION
**Pull request by moonyuet**
## Changelog Description
This PR is to implement validator for scene unknown nodes
Fixes https://github.com/ynput/ayon-maya/issues/228
Marked it as draft because the repair action might not be deletion of the node. Maybe perhaps saving the workfile scene could help.

## Testing notes:
1. Create any nodes(AAS, RSProxy) which is related to external renderers such as Arnold, Redshift etc.
2. Disable the external renderers
3. Create model with that node
4. Validation blocks if the unknown nodes are detected

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
+++ b/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
@@ -1,9 +1,9 @@
 from maya import cmds
 
 import os
+import inspect
 import pyblish.api
 from ayon_maya.api.action import SelectInvalidAction
-from ayon_core.pipeline import registered_host
 from ayon_core.pipeline.publish import (
     ValidateContentsOrder,
     RepairContextAction,
@@ -59,8 +59,7 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
             item["name"]: item["value"]
             for item in maya_settings["ext_mapping"]
         }
-        host = registered_host()
-        current_file = host.get_current_workfile()
+        current_file = context.data["currentFile"]
         extension = os.path.splitext(current_file)[-1].strip(".")
         correct_extension = ext_mapping.get(context.data["productBaseType"])
         return extension == correct_extension
@@ -83,7 +82,10 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
 
         invalid = self.get_invalid(context)
         if invalid:
-            raise PublishValidationError("Unknown nodes found: {0}".format(invalid))
+            raise PublishValidationError(
+                "Unknown nodes found: {0}".format(invalid),
+                description=self.get_description()
+            )
 
     @classmethod
     def repair(cls, context):
@@ -92,3 +94,17 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
                 force_delete(node)
             except RuntimeError as exc:
                 cls.log.error(exc)
+
+    def get_description(self) -> str:
+        return inspect.cleandoc("""
+            ## Unknown Nodes Found
+            Unknown nodes were found in the scene. This often happens if nodes from
+            plug-ins are used but are not available on this machine.
+            Note: Some studios use unknown nodes to store data on (as attributes)
+            because it's a lightweight node.
+            ### How to Fix
+            You can either:
+            - Install the missing plug-in that the unknown nodes belong to.
+            - Delete the unknown nodes from the scene. You can use the "Repair"
+            action to automatically delete the unknown nodes.
+        """)
--- a/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
+++ b/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
@@ -42,7 +42,7 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
     hosts = ['maya']
     families = ["model", "rig", "mayaScene", "look", "renderlayer", "yetiRig"]
     optional = True
-    label = "Unknown Nodes"
+    label = "Unknown Nodes in Scene"
     actions = [SelectInvalidAction, RepairContextAction]
 
     def _is_workfile_extension_align_with_extension_mapping(self, context) -> bool:
--- a/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
+++ b/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
@@ -73,10 +73,11 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
         if not self.is_active(context.data):
             return
 
-        if not self._is_workfile_extension_align_with_extension_mapping(context):
+        if self._is_workfile_extension_align_with_extension_mapping(context):
             self.log.warning(
-                "Workfile extension is not aligned with the extension mapping. "
-                "Skipping unknown nodes validation to prevent false positives."
+                "Workfile extension is not aligned with the extension mapping."
+                " Skipping unknown nodes validation to prevent false"
+                " positives."
             )
             return
 
--- a/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
+++ b/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
@@ -96,14 +96,18 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
             except RuntimeError as exc:
                 cls.log.error(exc)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: client/ayon_maya/plugins/publish/validate_unknown_nodes.py [hunk 1]
```diff
@@ -0,0 +1,94 @@
...
+        if not self._is_workfile_extension_align_with_extension_mapping(context):
+            self.log.warning(
+                "Workfile extension is not aligned with the extension mapping. "
+                "Skipping unknown nodes validation to prevent false positives."
+            )
+            return
+
+        invalid = self.get_invalid(context)
+        if invalid:
+            raise PublishValidationError("Unknown nodes found: {0}".format(invalid))
```
BigRoy: Can we add a description argument that explains the reasoning of the validator, etc.

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -0,0 +1,94 @@
...
+        Args:
+            context (pyblish.api.Context): The publish context.
+        """
+        maya_settings = context.data["project_settings"]["maya"]
+        ext_mapping = {
+            item["name"]: item["value"]
+            for item in maya_settings["ext_mapping"]
+        }
+        host = registered_host()
+        current_file = host.get_current_workfile()
```
BigRoy: We should be able to use just `context.data["currentFile"]` (if the file was saved; otherwise perhaps if not saved then return True always?)

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Args:
+        node (str): invalid node to delete
+    """
+    if cmds.objExists(node):
+        cmds.lockNode(node, lock=False)
+        cmds.delete(node)
+
+
+class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
```
BigRoy, line 27: Can we expose the enabled, optional, active state to settings please?

(seems I forgot to press submit on this before, sorry)

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        maya_settings = context.data["project_settings"]["maya"]
+        ext_mapping = {
+            item["name"]: item["value"]
+            for item in maya_settings["ext_mapping"]
+        }
+        current_file = context.data["currentFile"]
+        extension = os.path.splitext(current_file)[-1].strip(".")
+        correct_extension = ext_mapping.get(context.data["productBaseType"])
+        return extension == correct_extension
```
BigRoy: Context has no product base type, of course.
Also, unsaved file should not check (I suppose?)
```suggestion
        if not current_file:
            # Unsaved file
            return True

        workfile_extension = os.path.splitext(current_file)[-1].strip(".")
        for instance in context:
            # Skip inactivate instances
            if not instance.data.get("publish", True):
                continue

            # Consider only relevant instances
            if not pyblish.logic.plugins_by_families(
                [self.__class__],
                self.families
            ):
                continue

            instance_extension = ext_mapping.get(
                instance.data["productBaseType"]
            )
            if workfile_extension != instance_extension:
                return False
        return True
```

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -0,0 +1,110 @@
...
+    This differs from validate no unknown nodes since it checks the
+    full scene - not just the nodes in the instance.
+
+    """
+
+    order = ValidateContentsOrder
+    hosts = ['maya']
+    families = ["model", "rig", "mayaScene", "look", "renderlayer", "yetiRig"]
+    optional = True
+    label = "Unknown Nodes"
```
BigRoy: Differentiate label from other validator
```suggestion
    label = "Unknown Nodes in Scene"
```

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -0,0 +1,110 @@
...
+            Unknown nodes were found in the scene. This often happens if nodes from
+            plug-ins are used but are not available on this machine.
+            Note: Some studios use unknown nodes to store data on (as attributes)
+            because it's a lightweight node.
+            ### How to Fix
+            You can either:
+            - Install the missing plug-in that the unknown nodes belong to.
+            - Delete the unknown nodes from the scene. You can use the "Repair"
+            action to automatically delete the unknown nodes.
+        """)
```
BigRoy: Cosmetics
```suggestion
    @staticmethod
    def get_description() -> str:
        return inspect.cleandoc("""
            ## Unknown Nodes Found
            
            Unknown nodes were found in the scene. This often happens if nodes
            from plug-ins are used but are not available on this machine.
            Note: Some studios use unknown nodes to store data on (as 
            attributes) because it's a lightweight node.
            
            ### How to Fix
            
            You can either:
            - Install the missing plug-in that the unknown nodes belong to.
            - Delete the unknown nodes from the scene. You can use the "Repair"
            action to automatically delete the unknown nodes.
        """)
```

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -0,0 +1,110 @@
...
+        """Process all the nodes in the instance"""
+        if not self.is_active(context.data):
+            return
+
+        if not self._is_workfile_extension_align_with_extension_mapping(context):
+            self.log.warning(
+                "Workfile extension is not aligned with the extension mapping. "
+                "Skipping unknown nodes validation to prevent false positives."
+            )
+            return
```
BigRoy: This doesn't sound correct. The validation should check IF the extensions mismatch, so if it does NOT align? 🤔
Also, check line lengths with linter?
```suggestion
        if self._is_workfile_extension_align_with_extension_mapping(context):
            self.log.warning(
                "Workfile extension is not aligned with the extension mapping."
                " Skipping unknown nodes validation to prevent false"
                " positives."
            )
            return
```

BigRoy (changes requested):

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
BigRoy: Requires import `import pyblish.logic`

moonyuet: Fixed in [a448012](https://github.com/ynput/ayon-maya/pull/427/commits/a4480123b2a4b7b0e54badb326f4ac5065e3c807)

BigRoy (approved):
LGTM

LiborBatek (approved):
LGTM even tho it was harder to force it as `unknown nodes`  than just use e.g. arnold standin (as it got represented as empty group which triggered different ayon validator  :)

but when such unknown nodes present it correctly kicked in...

Image

PR #429, opened 2026-05-01
Add objExists guard to unlocked() context manager

rhdutta:
## Summary
- Fixes `ValueError: No object matches name: cameraMain` during camera publish to Unreal
- `unlocked()` context manager now checks `cmds.objExists(node)` before querying lock state
- `_default_update_instances()` skips imprinting when instance node no longer exists in scene

## Context
Camera publish to Unreal completes output successfully but fails during `_save_instance_changes` when the system attempts to query the lock state of a node that was deleted during extraction. The `bake_to_world_space` workflow creates temporary duplicate nodes for export, and by the time the post-publish save runs, the original instance objectSet may no longer exist.

## Changes
- `client/ayon_maya/api/lib.py` — `unlocked()` yields immediately if node doesn't exist
- `client/ayon_maya/api/plugin.py` — `_default_update_instances()` skips update for missing nodes
- `package.py` — Version bump to 0.4.19

## Tickets
- [ENG-4296](https://halon.atlassian.net/browse/ENG-4296) — Camera publish to Unreal not saving instance
- [ENG-4269](https://halon.atlassian.net/browse/ENG-4269) — Related duplicate

## Testing
- [ ] Camera publish to Unreal saves instance without error
- [x] Standard camera publish (non-Unreal) still works
- [x] Other publish types using `unlocked()` unaffected
- [ ] Verify instance data persists correctly after publish


package.py
```diff
@@ -1,14 +1,11 @@
 name = "maya"
 title = "Maya"
-version = "0.6.5+dev"
+version = "0.4.19"
```
moonyuet: can you revert this change?

LiborBatek:
As far as I can tell its working for me... I have tested with Camera, no any traces of that error in the script editor, also loaded into UE and tested with other publish type  (model wasnt affected by any means).

Not sure if I should use some custom product type / alias tho...so rather just commenting with my findings so far...

Let me know if I should test something specifically still.

package.py
rhdutta: Done. I also merged develop branch to avoid conflicts upstream.

rhdutta:
> Let me know if I should test something specifically still.

Should be good to go. Thank you!

client/ayon_maya/api/lib.py
```diff
@@ -120,6 +120,10 @@ def unlocked(node):
     Args:
         node (str): The name of the node to unlock.
     """
+    if not cmds.objExists(node):
+        yield
+        return
+
```
BigRoy: I actually think this is just confusing. It should really just error, we should just not be passing it nodes that do not exist.

BigRoy:
```suggestion
```

moonyuet: Agreed. The check on the `plugin.py` before this contextmanager function already did the same thing as the code.

BigRoy:
@rhdutta Please check if current fixes are good for you.

I'd actually argue that if the object does not exist when triggering `update` that we should perhaps log a warning at least - but maybe that's too verbose. 

How common is this, e.g. when does it happen?

BigRoy:
> when the system attempts to query the lock state of a node that was deleted during extraction

Why is extraction deleting that node to begin with? Seems like a totally different issue?

client/ayon_maya/api/plugin.py
```diff
@@ -309,6 +309,8 @@ def _default_update_instances(self, update_list):
         for created_inst, _changes in update_list:
             data = created_inst.data_to_store()
             node = data.get("instance_node")
+            if not node or not cmds.objExists(node):
+                continue
```
BigRoy, line 316:
```suggestion
                self.log.warning(
                    f"Skipping update for missing instance node: '{node}'"
                )
                continue
```

BigRoy (approved):
@moonyuet can you just confirm the code change - and merge if you're ok.

moonyuet (approved):
Tested successfully in Maya 2027

PR #432, opened 2026-05-11
Add active button for extractors in ayon setting 

moonyuet:
## Changelog Description
This PR is to add active button for extractors in ayon settings.
Resolve https://github.com/ynput/ayon-maya/issues/431

## Additional review information
Not included `ExtractLocalRender` as this is mandatory for the local rendering option.

## Testing notes:
1. Build and install addon
2. Test publishing all products if possible

PR #438, opened 2026-05-19
Fix: Scene extractor

antirotor:
## Changelog Description
Fix the bug introduced by #432 

Removed invalid mixin from class definition.

## Testing notes:
1) Create Maya Scene
2) Publish it successfuly


moonyuet (approved):
Tested successfully in Maya

BigRoy (approved):

PR #443, opened 2026-05-23
Export Pointcache: Write Face Sets: Disable forced shader assignments by default

### LINKED ISSUE #440: YN-0766: Extract Pointcache: force shader assignment only when writing face sets

When Extract Pointcache trigger it always does the forced shader assignment because `instance.data["writeFaceSets"]` isn't actually set only when write face sets is enabled. As such it's wrongly if branched here: https://github.com/ynput/ayon-maya/blob/develop/client/ayon_maya/plugins/publish/extract_pointcache.py#L245

&nbsp;

Also, the forcing should really be optional behavior so expose it in settings. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1504904487465128109)

_This issue was automatically created from AYON ticket [YN-0766]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=ed76a89056d711f18c38db493397c2f2)

BigRoy:
## Changelog Description

Make the forced face shader assignment optional (for the studio admin) to allow disabling/enabling that feature. Also, disable it by default now because the behavior is rather special and non-standard for Alembic export behavior.

To preserve the original behavior, make sure to enable the `writeFaceSetsForceFaceAssignment` flag in settings

This also fixes that if this setting would be enabled, but writing face sets would be disabled then it should not trigger - whereas previously the forced face shader assignment _always_ triggered.

Fix https://github.com/ynput/ayon-maya/issues/440

## Testing notes:
1. Extract Pointcache Write face sets should work with that flag enabled in settings, but with `writeFaceSetsForceFaceAssignment` disabled.
2. Extract Pointcache Write face sets should work with that flag enabled in settings, but with `writeFaceSetsForceFaceAssignment` enabled.
3. Different in resulting file should be that with it enabled any non-face assignments should've still ben turned into face sets on write.

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_pointcache.py
+++ b/client/ayon_maya/plugins/publish/extract_pointcache.py
@@ -68,6 +68,7 @@ class ExtractAlembic(plugin.MayaExtractorPlugin,
     writeColorSets = False
     writeCreases = False
     writeFaceSets = False
+    writeFaceSetsForceFaceAssignment = False
     writeNormals = True
     writeUVSets = False
     writeVisibility = False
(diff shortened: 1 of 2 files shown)


moonyuet (approved):
Tested successfully in maya 2027

PR #449, opened 2026-06-17
Renderlayer should not error when collection is under the group and overrides have setup 

### LINKED ISSUE #195: AY-7247_Get recursive overrides not using groups

Problem when collecting overrides in Maya is that maya.app.renderSetup.model.utils.getOverridesRecursive(), which is the one used in lib_rendersetup.py,  is recursive only at a collection level. When using also groups these overrides are not considered.

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/768872686549205012/1316040613069848667)

_This issue was automatically created from Clickup ticket [AY-7247](https://app.clickup.com/t/86c19vm6h)_

moonyuet:
## Changelog Description
This PR is to make sure renderlayer does not result any error whe collection is under the group and overrides have been created.
Fixes: https://github.com/ynput/ayon-maya/issues/195

n/a

## Testing notes:
1. Create a renderlayer
2. Have the collection with absolute overrides and parented it to the group under renderlayer
3. Have the collection with absolute override under renderlayer(without group).
4. Open Publisher
5. Publish
6. Should not error any validator(except the validation error) and published successfully to farm

### DIFF
--- a/client/ayon_maya/api/lib_rendersetup.py
+++ b/client/ayon_maya/api/lib_rendersetup.py
@@ -364,28 +364,29 @@ def iter_layer_overrides(layer):
     Yields:
         Override: Each override object found in the layer hierarchy.
     """
-    queue = [layer]
-    # layer.getChildren() does not return groups even though
-    # getChildren on groups or collections does return them
-    if hasattr(layer, "getChildren"):
-        queue.extend(layer.getChildren())
+    stack = []
+    # Seed stack with both the layer's children and groups (layer.getChildren()
+    # omits groups, unlike groups/collections themselves).
     if hasattr(layer, "getGroups"):
-        queue.extend([g for g in layer.getGroups() if g.isEnabled()])
+        stack.extend(reversed([g for g in layer.getGroups() if g.isEnabled()]))
+    if hasattr(layer, "getChildren"):
+        # Pushed last so it's visited first when popping from the stack.
+        stack.extend(reversed(layer.getChildren()))
 
-    for obj in queue:
+    while stack:
+        obj = stack.pop()
         if isinstance(obj, Override):
             yield obj
-        else:
-            if hasattr(obj, "isRenderable") and not obj.isRenderable():
-                # Skip disabled groups/collections as their overrides are not
-                # active
-                continue
-            if hasattr(obj, "isEnabled") and not obj.isEnabled():
-                continue
+            continue
 
-            if hasattr(obj, "getChildren"):
-                queue.extend(obj.getChildren())
+        if hasattr(obj, "isRenderable") and not obj.isRenderable():
+            # Skip disabled groups/collections as their overrides are not active
+            continue
+        if hasattr(obj, "isEnabled") and not obj.isEnabled():
+            continue
 
+        if hasattr(obj, "getChildren"):
+            stack.extend(reversed(obj.getChildren()))
 
 def get_shader_in_layer(node, layer):
     """Return the assigned shader in a renderlayer without switching layers.
--- a/client/ayon_maya/api/lib_rendersetup.py
+++ b/client/ayon_maya/api/lib_rendersetup.py
@@ -14,6 +14,7 @@ import logging
 import maya.app.renderSetup.model.utils as utils
 from maya.app.renderSetup.model import renderSetup
 from maya.app.renderSetup.model.override import (
+    Override,
     AbsOverride,
     RelOverride,
     UniqueOverride
@@ -299,7 +300,7 @@ def get_attr_overrides(node_attr, layer,
     # Iterate over the overrides in reverse so we get the last
     # overrides first and can "break" whenever an absolute
     # override is reached
-    layer_overrides = list(utils.getOverridesRecursive(rs_layer))
+    layer_overrides = list(iter_layer_overrides(rs_layer))
     for layer_override in reversed(layer_overrides):
 
         if skip_disabled and not layer_override.isEnabled():
@@ -347,6 +348,23 @@ def get_attr_overrides(node_attr, layer,
             break
 
     return reversed(plug_overrides)
+
+
+def iter_layer_overrides(layer):
+    """Iterate layer overrides.
+
+    Args:
+        layer (RenderLayer): RenderLayer to iterate the overrides for
+
+    Yields:
+        Override: Each override object found in the layer hierarchy.
+    """
+    queue = [layer]
+    for obj in queue:
+        if isinstance(obj, Override):
+            yield obj
+        else:
+            queue.extend(obj.getChildren())
 
 
 def get_shader_in_layer(node, layer):
--- a/client/ayon_maya/api/lib_rendersetup.py
+++ b/client/ayon_maya/api/lib_rendersetup.py
@@ -11,7 +11,6 @@ from maya import cmds
 import maya.api.OpenMaya as om
 import logging
 
-import maya.app.renderSetup.model.utils as utils
 from maya.app.renderSetup.model import renderSetup
 from maya.app.renderSetup.model.override import (
     Override,
--- a/client/ayon_maya/api/lib_rendersetup.py
+++ b/client/ayon_maya/api/lib_rendersetup.py
@@ -367,16 +367,24 @@ def iter_layer_overrides(layer):
     queue = [layer]
     # layer.getChildren() does not return groups even though
     # getChildren on groups or collections does return them
-    queue.extend(layer.getGroups())
+    if hasattr(layer, "getChildren"):
+        queue.extend(layer.getChildren())
+    if hasattr(layer, "getGroups"):
+        queue.extend([g for g in layer.getGroups() if g.isEnabled()])
+
     for obj in queue:
         if isinstance(obj, Override):
             yield obj
         else:
-            if not obj.isEnabled():
+            if hasattr(obj, "isRenderable") and not obj.isRenderable():
                 # Skip disabled groups/collections as their overrides are not
                 # active
                 continue
-            queue.extend(obj.getChildren())
+            if hasattr(obj, "isEnabled") and not obj.isEnabled():
+                continue
+
+            if hasattr(obj, "getChildren"):
+                queue.extend(obj.getChildren())
 
 
 def get_shader_in_layer(node, layer):
--- a/client/ayon_maya/api/lib_rendersetup.py
+++ b/client/ayon_maya/api/lib_rendersetup.py
@@ -351,7 +351,13 @@ def get_attr_overrides(node_attr, layer,
 
 def iter_layer_overrides(layer):
     """Iterate layer overrides.
-
+    Note: We cannot use `maya.app.renderSetup.model.utils.getOverridesRecursive`
+        here because it traverses via `getChildren()` which, on a RenderLayer
+        (unlike on a Group or Collection), does NOT return groups. That means
+        any overrides living inside groups would be silently skipped.
+        We work around this by seeding the queue with both `getChildren()` and
+        `getGroups()` for the top-level layer only; from that point on,
+        `getChildren()` on groups/collections correctly returns nested items.
     Args:
         layer (RenderLayer): RenderLayer to iterate the overrides for
 
@@ -359,10 +365,17 @@ def iter_layer_overrides(layer):
         Override: Each override object found in the layer hierarchy.
     """
     queue = [layer]
+    # layer.getChildren() does not return groups even though
+    # getChildren on groups or collections does return them
+    queue.extend(layer.getGroups())
     for obj in queue:
         if isinstance(obj, Override):
             yield obj
         else:
+            if not obj.isEnabled():
+                # Skip disabled groups/collections as their overrides are not
+                # active
+                continue
             queue.extend(obj.getChildren())
 
 


client/ayon_maya/api/lib_rendersetup.py
```diff
@@ -349,6 +349,23 @@ def iter_override_targets(override):
...
+def iter_layer_overrides(layer):
+    """Iterate layer overrides.
+
+    Args:
+        layer (RenderLayer): RenderLayer to iterate the overrides for
+
+    Yields:
+        Override: Each override object found in the layer hierarchy.
+    """
+    queue = [layer]
```
BigRoy: I think this also needs the `getGroups()` call because `layer.getChildren()` I believe excludes the groups?

See internal note [here](https://discord.com/channels/517362899170230292/768872686549205012/1316353984591564853).

Also, please confirm/test this question:
> However, I'm not sure if we should technically still also be checking whether any of its parents is disabled, or whether override.isEnabled() also considers those?

client/ayon_maya/api/lib_rendersetup.py
```diff
@@ -349,6 +349,23 @@ def iter_override_targets(override):
     return reversed(plug_overrides)
 
 
+def iter_layer_overrides(layer):
+    """Iterate layer overrides.
+
```
BigRoy: Might be good to explain why we can't use `maya.app.renderSetup.model.utils.getOverridesRecursive` for our future selves?

client/ayon_maya/api/lib_rendersetup.py
```diff
@@ -349,6 +349,36 @@ def iter_override_targets(override):
...
+    queue.extend(layer.getGroups())
+    for obj in queue:
+        if isinstance(obj, Override):
+            yield obj
+        else:
+            if not obj.isEnabled():
+                # Skip disabled groups/collections as their overrides are not
+                # active
+                continue
+            queue.extend(obj.getChildren())
```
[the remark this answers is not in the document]
BigRoy: This actually sounds like there could be a source of truth here - definitely something to investigate. I have a gut feeling that if this is the case, that the proposed code change here doesn't technically fully accurately resolve it. Because technically it's also the case that overrides are ordered, and I believe groups can be ordered in-between children overrides directly - as such, stacking groups at the end of current children may technically be wrong too.

I'm also, still very confused why `getOverridesRecursive` would skip groups in the first place 🤷‍♂️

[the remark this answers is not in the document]
moonyuet: I got an error when the code includes `queue.extend(layer.getGroups())`. Removing the line would end up the same result with `getOverridesRecursive`.
```python
from maya.app.renderSetup.model import renderSetup
from maya.app.renderSetup.model.override import Override


def iter_layer_overrides(layer):
    queue = [layer]
    
    # layer.getChildren() does not return groups even though
    # getChildren on groups or collections does return them
    
    for obj in queue:
        if isinstance(obj, Override):
            yield obj
        else:
            queue.extend(obj.getChildren())


rs = renderSetup.instance()
for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override:", override.name())

# found override: absOverride2
# Finished...
# found override: absOverride2
```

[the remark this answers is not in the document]
moonyuet: If we fixed the `layer.getGroups` it should be also getting the overrides
```python
from maya.app.renderSetup.model import renderSetup
from maya.app.renderSetup.model.override import Override


def iter_layer_overrides(layer):
    queue = [layer]
    
    # layer.getChildren() does not return groups even though
    # getChildren on groups or collections does return them
    
    for obj in queue:
        if isinstance(obj, Override):
            yield obj
        else:
            queue.extend(obj.getChildren())


rs = renderSetup.instance()
for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs)):
    print("found override:", override.name())

found override: absOverride2
Finished...
found override: absOverride2
from maya.app.renderSetup.model import renderSetup
from maya.app.renderSetup.model.override import Override


def iter_layer_overrides(layer):
    queue = [layer]
    queue.extend(layer.getGroups() if hasattr(layer, "getGroups") else [])
    # layer.getChildren() does not return groups even though
    # getChildren on groups or collections does return them
    
    for obj in queue:
        if isinstance(obj, Override):
            yield obj
        else:
            queue.extend(obj.getChildren())


rs = renderSetup.instance()
for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override:", override.name())

# found override: absOverride2
# Finished...
# found override: absOverride1
# Finished...
# found override: absOverride2

**Comment by moonyuet** [comment]
@BigRoy I tested with some edge cases with the code snippet below and come up with [41907a2](https://github.com/ynput/ayon-maya/pull/449/commits/41907a232c8d6b1d10d4e5e1874960f27abe2c70) 

```python
from maya.app.renderSetup.model import renderSetup
from maya.app.renderSetup.model.override import Override


def iter_layer_overrides(layer):
    queue = [layer]
    # layer.getChildren() does not return groups even though
    # getChildren on groups or collections does return them
    if hasattr(layer, "getGroups"):
        for layer_group in layer.getGroups():
            if layer_group.isEnabled():
                queue.append(layer_group)

    for obj in queue:
        if isinstance(obj, Override):
            yield obj
        else:
            if hasattr(obj, "isRenderable") and not obj.isRenderable():
                # Skip disabled groups/collections as their overrides are not
                # active
                continue
            if hasattr(obj, "isEnabled") and not obj.isEnabled():
                continue
            queue.extend(obj.getChildren())
rs = renderSetup.instance()

for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override:", override.name())
# found override: absOverride2
# Finished...
# found override: absOverride1
# Finished...
# found override: absOverride3
# Finished...

# for the iteration from list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())
# found override: absOverride2
# found override: absOverride3
```
This is the sample scene I tested. 
[test_renderlayer_scene.zip](https://github.com/user-attachments/files/29157200/test_renderlayer_scene.zip)

image

LiborBatek (approved):
I have  followed the testing steps and all been functional when using `layers` and `groups` with overrides.

LGTM

client/ayon_maya/api/lib_rendersetup.py
```diff
@@ -349,6 +349,44 @@ def iter_override_targets(override):
...
+    Yields:
+        Override: Each override object found in the layer hierarchy.
+    """
+    queue = [layer]
+    # layer.getChildren() does not return groups even though
+    # getChildren on groups or collections does return them
+    if hasattr(layer, "getChildren"):
+        queue.extend(layer.getChildren())
+    if hasattr(layer, "getGroups"):
+        queue.extend([g for g in layer.getGroups() if g.isEnabled()])
```
BigRoy: I'm a bit worried about the ordering here. Because technically, in this scenario:

```
layer
  - group
    - override
  - override
  - group
    - overide
  - override
```

Let's for the sake of it assume that each of these overrides affect the same attribute.

According to Maya docs:

> Collections are interpreted top-down. Therefore, if two (or more) of your collections share objects, and one of your collections is enabled and the other is disabled, it is the collection furthest down in the list that determines whether the shared object is included in the layer

- https://help.autodesk.com/view/MAYAUL/2017/ENU/?guid=GUID-646154F9-E4D2-4A3E-A0D9-B2F062960A4E
- https://help.autodesk.com/view/MAYAUL/2017/ENU/?guid=GUID-2B94C5A1-70DE-4280-BB44-9370C5B778A5

As such, if we just happen to queue groups after children (or reversed) we're losing the ordering between them and with that totally lose the 'ordering'.

This issue actually already comes from the `getChildren` call on its own as well, because if you'd have deeper nesting - than the deeper nested ones would appear later in the queue (because we queue breadth-first) and hence suddenly what's deeper in a hierarchy may turn to be a stronger override perhaps?

I believe the original function we called would retrieve solely the resulting applying override - not the ones that appeared earlier so that anything that called `utils.getOverridesRecursive` would get only the ones we care about applying. But if not, we may have actually been computing the overrides all wrong to begin with? 🤔 

Can we prepare some of these test cases? Where we just start making a few layers, where each layers essentially has this multiple sort of overrides to a single attribute?

```
layer
  - coll/coll/coll/override
  - coll/override <- I suspect this should be applied? (check value in the layer)
```

```
layer
  - coll/override
  - coll/coll/coll/override <- I suspect this should be applied? (check value in the layer)
```

```
layer
    - coll/override
    - group/coll/override
    - coll/override <- I suspect this should be applied? (check value in the layer)
```

```
layer
    - group/coll/override
    - coll/override
    - group/coll/override <- I suspect this should be applied? (check value in the layer)
```

etc.

moonyuet: I tested with four scenarios you suggested. And the ordering of the overrides might not be correct but my function still gets all the overrides in the visible layers: (I used the function https://github.com/ynput/ayon-maya/pull/449#issuecomment-4756750581above) and utils.getOverridesRecrusive.
The first scenarios. 
```
for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override with override function:", override.name())

found override: absOverride9
Finished...
found override: absOverride8
Finished...
found override with override function: absOverride8
found override with override function: absOverride9
```
The second scenario:
```
for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override with override function:", override.name())

found override: absOverride10
Finished...
found override: absOverride4
Finished...
found override with override function: absOverride10
found override with override function: absOverride4
``` 
Third scanerio:
```
rs = renderSetup.instance()

for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override with override function:", override.name())

found override: absOverride6
Finished...
found override: absOverride12
Finished...
found override: absOverride11
Finished...
found override with override function: absOverride6
found override with override function: absOverride12
```
Fourth scanerio
```
for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override with override function:", override.name())

found override: absOverride13
Finished...
found override: absOverride7
Finished...
found override: absOverride14
Finished...
found override with override function: absOverride13
```
So maybe the next step we can think of how to deal make sure the overrides at the correct orders.

BigRoy: I think what we can best do is confirm the order that Maya actually applies them and make sure we result in the same final result. We may as well even implement tests to ensure it's always the case - either by having an example render setup json file that we just "import" in a test file, do the check from the main layer, then switch to the relevant layer and see if the values indeed align with out computed ones, etc.?

moonyuet:
I committed the code suggestion from copilot and all scenarios work as expected.
@BigRoy I used the script below
```
def iter_layer_overrides(layer):
    queue = [layer]
    # layer.getChildren() does not return groups even though
    # getChildren on groups or collections does return them
    if hasattr(layer, "getGroups"):
        for layer_group in layer.getGroups():
            if layer_group.isEnabled():
                queue.append(layer_group)

    for obj in queue:
        if isinstance(obj, Override):
            yield obj
        else:
            if hasattr(obj, "isRenderable") and not obj.isRenderable():
                # Skip disabled groups/collections as their overrides are not
                # active
                continue
            if hasattr(obj, "isEnabled") and not obj.isEnabled():
                continue
            queue.extend(obj.getChildren())
rs = renderSetup.instance()

for override in iter_layer_overrides(rs.getVisibleRenderLayer()):
    
    print("found override:", override.name())
    print("Finished...")
import maya.app.renderSetup.model.utils as utils
for override in list(utils.getOverridesRecursive(rs.getVisibleRenderLayer())):
    print("found override with override function:", override.name())
```
log
```
found override: absOverride2
Finished...
found override: absOverride1
Finished...
found override: absOverride3
Finished...
found override with override function: absOverride2
found override with override function: absOverride3
```
```
found override: absOverride9
Finished...
found override: absOverride8
Finished...
found override with override function: absOverride8
found override with override function: absOverride9
```
```
found override: absOverride10
Finished...
found override: absOverride4
Finished...
found override with override function: absOverride10
found override with override function: absOverride4
```
```
found override: absOverride6
Finished...
found override: absOverride12
Finished...
found override: absOverride11
Finished...
found override with override function: absOverride6
found override with override function: absOverride12
```
```

found override: absOverride13
Finished...
found override: absOverride7
Finished...
found override: absOverride14
Finished...
found override with override function: absOverride13
```
Should we add the sorted to make sure the overrides at correct orders?

moonyuet:
@BigRoy can you review? Thanks!
I tested with different scenarios as comments shown above

PR #450, opened 2026-06-20
Improve the logic for Validate Unknown Nodes for readability

moonyuet:
## Changelog Description
This PR is to improve the logic for Validate Unknown Nodes for readability.

## Testing notes:
1. Launch Maya with some plugins disabled such as RS, Vray
2. Publish
3. Validation blocked.

### DIFF
--- a/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
+++ b/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
@@ -54,6 +54,10 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
 
         Args:
             context (pyblish.api.Context): The publish context.
+.
+        Returns:
+            bool: True if the workfile extension is aligned with the extension
+                mapping, False otherwise.
         """
         maya_settings = context.data["project_settings"]["maya"]
         ext_mapping = {
@@ -62,7 +66,7 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
         }
         current_file = context.data["currentFile"]
         if not current_file:
-            # Unsaved file
+            # Unsaved file: do not block validation
             return True
         workfile_extension = os.path.splitext(current_file)[-1].strip(".")
         for instance in context:
@@ -78,9 +82,9 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
             instance_extension = ext_mapping.get(
                 instance.data["productBaseType"]
             )
-            if workfile_extension != instance_extension:
-                return False
-        return True
+            if workfile_extension == instance_extension:
+                return True
+        return False
 
     @staticmethod
     def get_invalid(context) -> list:
@@ -91,7 +95,7 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
         if not self.is_active(context.data):
             return
 
-        if self._is_workfile_extension_align_with_extension_mapping(context):
+        if not self._is_workfile_extension_align_with_extension_mapping(context):
             self.log.warning(
                 "Workfile extension is not aligned with the extension mapping."
                 " Skipping unknown nodes validation to prevent false"
--- a/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
+++ b/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
@@ -24,6 +24,12 @@ def force_delete(node: str) -> None:
         cmds.delete(node)
 
 
+class RemoveUnknownNodesAction(RepairContextAction):
+    """Repair action to remove unknown nodes from the scene."""
+
+    label = "Remove Unknown Nodes"
+
+
 class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
                                 OptionalPyblishPluginMixin):
     """Checks to see if there are any unknown nodes in the scene.
@@ -44,7 +50,7 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
     families = ["model", "rig", "mayaScene", "look", "renderlayer", "yetiRig"]
     optional = True
     label = "Unknown Nodes in Scene"
-    actions = [SelectInvalidAction, RepairContextAction]
+    actions = [SelectInvalidAction, RemoveUnknownNodesAction]
 
     def _is_workfile_extension_align_with_extension_mapping(self, context) -> bool:
         """Check if the workfile extension is aligned with the extension mapping.
@@ -134,6 +140,6 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
 
             You can either:
             - Install the missing plug-in that the unknown nodes belong to.
-            - Delete the unknown nodes from the scene. You can use the "Repair"
-            action to automatically delete the unknown nodes.
+            - Delete the unknown nodes from the scene. You can use the "Remove
+            Unknown Nodes" action to automatically delete the unknown nodes.
         """)
--- a/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
+++ b/client/ayon_maya/plugins/publish/validate_unknown_nodes.py
@@ -100,7 +100,8 @@ class ValidateSceneUnknownNodes(pyblish.api.ContextPlugin,
 
         if not self._is_workfile_extension_align_with_extension_mapping(context):
             self.log.warning(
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
LGTM - I wonder if it's worth for Libor to do a test review of this?

LiborBatek (commented):
It does show up as an optional validator in the Publisher on the context widget which is good.

Im not just fully convinced by the message it shows as my scene extension and mapping actually matches even tho the info message says something different. 

Image


Image

and my workfile and extension:

Image


Am I getting it right or not really?!

I guess maybe just leaving out the first info sentence might be okeish and just keeping rest of the message?

moonyuet:
> It does show up as an optional validator in the Publisher on the context widget which is good.
> 
> Im not just fully convinced by the message it shows as my scene extension and mapping actually matches even tho the info message says something different.
> 
> Image Image
> and my workfile and extension:
> 
> Image
> Am I getting it right or not really?!
> 
> I guess maybe just leaving out the first info sentence might be okeish and just keeping rest of the message?

You are right, and you are not the only one to mention this. See https://discord.com/channels/517362899170230292/563751989075378201/1521431127124873296. yes I agree that we should remove the first info sentence.

LiborBatek (changes requested):
While now the message is not referring the extension mapping mismatch which is good....I have faced quite major issue on other front...

Even tho having any available validators for `unknown nodes` set as enabled I was able to pass validations and try to publish which obviously failed due to missing addon loaded...so there is def. something wrong with those validators...

or am I missing something here? I guess its not working atm...or?


maya script editor proving Im missing Yeti addon and having unknown nodes in the scene
Image

while still validating ok as Im having those validators enabled

Image

went through without worries:

Image

but failed and with a good reasoning (missing yeti addon)

Image

@moonyuet ideas?

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -84,18 +88,19 @@ def _is_workfile_extension_align_with_extension_mapping(self, context) -> bool:
...
             return
 
-        if self._is_workfile_extension_align_with_extension_mapping(context):
+        if not self._is_workfile_extension_align_with_extension_mapping(context):
             self.log.warning(
-                "Workfile extension is not aligned with the extension mapping."
-                " Skipping unknown nodes validation to prevent false"
-                " positives."
+                "Skipping unknown nodes validation to prevent false positives."
             )
```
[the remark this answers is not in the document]
moonyuet: Fixed in [fd9df19](https://github.com/ynput/ayon-maya/pull/450/commits/fd9df19cb5948d4cec79006dd384d70b6e7d0960)

LiborBatek (changes requested):
So now it correctly kicks in and validations been triggered when having some `unknown nodes` in the workfile which is good!

I would just tweak the `Repair` term to `Remove` in this particular case as the repairing means just removing those unknown nodes which is a destructive action so `repair` is not the proper name for this action... as seen on the scrn below

Image

And another thing is I have still experienced in one scene the warning about extension mismatch which is weird as they should match and be `.ma` both the workfile and the `ayon-maya` preference for saving...could you inspect it one more time if the logic been correct...or maybe even reformulating the info message? Not sure maybe @BigRoy can give some ideas...

Image

moonyuet:
> So now it correctly kicks in and validations been triggered when having some `unknown nodes` in the workfile which is good!
> 
> I would just tweak the `Repair` term to `Remove` in this particular case as the repairing means just removing those unknown nodes which is a destructive action so `repair` is not the proper name for this action... as seen on the scrn below
> 
> Image

Fixed in [baabbdc](https://github.com/ynput/ayon-maya/pull/450/commits/baabbdc172f819b583c7ebbcae225bb0d5a85e92)
image

> And another thing is I have still experienced in one scene the warning about extension mismatch which is weird as they should match and be `.ma` both the workfile and the `ayon-maya` preference for saving...could you inspect it one more time if the logic been correct...or maybe even reformulating the info message? Not sure maybe @BigRoy can give some ideas...
> 
> Image

I think the best approach is to filter it by instance like.
```python
for instance in context:
     if self.__is_workfile_extension_align_with_extension_mapping(instance):
        self.log.warning(""Skipping unknown nodes validation because the workfile extension "....)
		return
```

LiborBatek (approved):
Now all working ok and instead of Repair action there is a clear declaration of `Remove unknown nodes` LGTM.

Good to go I guess...

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -84,18 +94,20 @@ def _is_workfile_extension_align_with_extension_mapping(self, context) -> bool:
 
     @staticmethod
     def get_invalid(context) -> list:
-        return cmds.ls(type="unknown")
+        return (
+            (cmds.ls(type="unknown") or [])
+            + (cmds.ls(type="unknownDag") or [])
+        )
```
BigRoy: Should probably be this:

1. You can request multiple types at once.
2. If it's a DAG node, you'll want the full paths.

```suggestion
        return cmds.ls(type=("unknown", "unknownDag", long=True)
```

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -84,18 +94,20 @@ def _is_workfile_extension_align_with_extension_mapping(self, context) -> bool:
...
 
-        if self._is_workfile_extension_align_with_extension_mapping(context):
+        if not self._is_workfile_extension_align_with_extension_mapping(context):
             self.log.warning(
-                "Workfile extension is not aligned with the extension mapping."
-                " Skipping unknown nodes validation to prevent false"
-                " positives."
+                "Skipping unknown nodes validation because the workfile extension "
+                "does not match extension mapping; this prevents false positives."
             )
```
BigRoy: Is the log message incorrect here?

When is this validation relevant in practice? If I understand correctly it's only relevant if we're about to export to a maya scene extension that is NOT the current extension (because maya disallows that when there are unknown nodes.)

As such, it really means:
- If one of the publishes outputs an extension that doesn't match the current extension then the validator should kick in and validate things.

Right @LiborBatek @moonyuet ?

BigRoy (commented):
Some notes - forgot to press "Submit Review" apparently.

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
moonyuet: Yes it is.

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
moonyuet: [06bd021](https://github.com/ynput/ayon-maya/pull/450/commits/06bd02153042efeeb91154612df7460b11e9c21d)

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
BigRoy: Then the log message is the opposite of what it should be saying - and it should not be a warning?

How about:

info or debug log:

> Allowing unknown nodes because file extension matches publish file export extensions. Skipping unknown nodes validation.


And still debug log the unknown nodes it found:

> Found 12 unknown nodes: node1, node2, node3

Does that make sense?

@LiborBatek what do you think?

LiborBatek:
@BigRoy yeah that sounds good.. I was mostly complaining about the fact my extensions been matching and the original message was confusing as it suggested they were not matching (which was not the case) so this proposal from you sounds good to me!

client/ayon_maya/plugins/publish/validate_unknown_nodes.py
```diff
@@ -84,23 +94,23 @@ def _is_workfile_extension_align_with_extension_mapping(self, context) -> bool:
...
-            self.log.warning(
-                "Workfile extension is not aligned with the extension mapping."
-                " Skipping unknown nodes validation to prevent false"
-                " positives."
-            )
-            return
-
         invalid = self.get_invalid(context)
         if invalid:
+            self.log.debug("Unknown {0} nodes found: {1}".format(len(invalid), invalid))
```
BigRoy:
```suggestion
            self.log.debug(f"Unknown {len(invalid)} nodes found: {invalid}")
```

BigRoy (approved):
Nice. Just formatting cosmetics to note still.

PR #457, opened 2026-07-13
ReferenceLoader: Update Namespace for switching container

### LINKED ISSUE #14: Switch Container: Update namespaces + change animation publish instances

There are cases where switching a product for a container (loaded instance) starts to become very confusing because the namespaces by default include the folder and product name which when switching does not follow along.

The same goes for switching rigs, the publish instances keep their old product names - which in many cases adds to the confusion if those for example are still named like the old folder they were maybe loaded from.

Add an option or setting to allow the namespace and animation instances to be updated along when switching the container.

Change the namespace and publish instances manually - but this is a lot of work usually.

### Additional context:

Traditionally namespaces in Maya were left alone for a few reasons:

- They used to be buggy. Reference edits often got disconnected because Maya for whatever reason still stored them as edits applied with the original namespace.
- We used to publish a lot with strip namespaces OFF, which means that in the lighting render scenes you may be dependent on the namespace to remain the same to have any edits applied (e.g. materials assigned) to be preserved edits when loading a new e.g. poincache or animation instance which internally suddenly would have different namespaces. However, I believe for some time now strip namespaces ON has been the default behavior (if not, it maybe should be.)

moonyuet:
## Changelog Description
This PR is to add support to update Namespace for switching container
Resolve https://github.com/ynput/ayon-maya/issues/14
The method needs some discussion as it may not be safe to do so.
image
Test with loading rig and see if the product name of the animation instance has been switched as the namespace switched.
Test with the general load(e.g. model, pointcache) and switch assets and see if the namespace has been updated. 

## Testing notes:
1. Turn on `ayon+settings://maya/load/reference_loader/update_namespace_on_switch`
2. Load asset
3. Switch Asset
4. Namespace should be updated accordingly

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -121,14 +121,16 @@ def unlocked(node):
         node (str): The name of the node to unlock.
     """
     has_locked = cmds.lockNode(node, query=True, lock=True)[0]
-    cmds.lockNode(node, lock=False)
-
+    node_uuid = cmds.ls(node, uuid=True)[0]
     try:
+        cmds.lockNode(node, lock=False)
         yield
 
     finally:
-        if cmds.objExists(node):
-            cmds.lockNode(node, lock=has_locked)
+        if not cmds.objExists(node):
+            # Try and find the node by uuid
+            node = cmds.ls(node_uuid)[0]
+        cmds.lockNode(node, lock=has_locked)
 
 
 @contextlib.contextmanager
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4373,10 +4373,6 @@ def update_animation_instance(
     )
 
     for object_set in object_sets:
-        # Only consider empty object sets
-        members = cmds.sets(object_set, query=True)
-        if members:
-            continue
         # Ignore referenced object sets
         if cmds.referenceQuery(object_set, isNodeReferenced=True):
             continue
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4341,6 +4341,91 @@ def get_creator_identifier(node: str) -> str | None:
     return get_attribute(f"{node}.creator_identifier")
 
 
+def update_animation_instance(
+    container: dict,
+    context: dict,
+    namespace: str
+) -> None:
+    """Update the name of animation instance
+
+    Args:
+        container (dict): container
+        context (dict): context
+        namespace (str): namespace
+    """
+    members = get_container_members(container)
+    object_sets = set()
+    for member in members:
+        object_sets.update(
+            cmds.listSets(object=member, extendToShape=False) or []
+        )
+    object_sets = cmds.ls(object_sets, type="objectSet")
+    create_context = CreateContext(registered_host(), discover_publish_plugins=False)
+    animation_creator_identifier = "io.openpype.creators.maya.animation"
+    animation_creator = create_context.creators.get(animation_creator_identifier)
+    variant = get_rig_animation_instance_variant(context, namespace)
+    target_name = animation_creator.get_product_name(
+        project_name=create_context.get_current_project_name(),
+        folder_entity=create_context.get_current_folder_entity(),
+        task_entity=create_context.get_current_task_entity(),
+        variant=variant,
+        instance=None
+    )
+
+    for object_set in object_sets:
+        # Only consider empty object sets
+        members = cmds.sets(object_set, query=True)
+        if members:
+            continue
+        # Ignore referenced object sets
+        if cmds.referenceQuery(object_set, isNodeReferenced=True):
+            continue
+        # Then only here confirm whether this is an animation instance, if so
+        # then we will want to auto-remove the instance
+        if get_creator_identifier(object_set) == animation_creator_identifier:
+            new_objectset = cmds.rename(object_set, target_name)
+            set_attribute(node=new_objectset, attribute="productName", value=new_objectset)
+
+
+def get_rig_animation_instance_variant(context, namespace, options=None)-> str:
+    folder_entity = context["folder"]
+    product_entity = context["product"]
+    product_type = product_entity["productType"]
+    product_base_type = product_entity.get("productBaseType")
+    if not product_base_type:
+        product_base_type = product_type
+    product_name = product_entity["name"]
+
+    custom_product_name = options.get("animationProductName")
+    if custom_product_name:
+        for old_key, new_key in (
+            ("{family}", "{product[basetype]}"),
+            ("{asset}", "{folder[name]}"),
+            ("{subset}", "asset"),
+        ):
+            if old_key in custom_product_name:
+                log.warning(f"Using deprecated template key '{old_key}'")
+                custom_product_name = custom_product_name.replace(
+                    old_key, new_key
+                )
+
+        formatting_data = {
+            "folder": {
+                "name": folder_entity["name"]
+            },
+            "product": {
+                "name": product_name,
+                "type": product_type,
+                "basetype": product_base_type,
+            },
+        }
+        return get_custom_namespace(
+            custom_product_name.format(**formatting_data)
+        )
+
+    return namespace
+
+
 def create_rig_animation_instance(
     nodes, context, namespace, options=None, log=None
 ):
@@ -4389,40 +4474,7 @@ def create_rig_animation_instance(
     )
     assert roots, "No root nodes in rig, this is a bug."
 
-    folder_entity = context["folder"]
-    product_entity = context["product"]
-    product_type = product_entity["productType"]
-    product_base_type = product_entity.get("productBaseType")
-    if not product_base_type:
-        product_base_type = product_type
-    product_name = product_entity["name"]
-
-    custom_product_name = options.get("animationProductName")
-    if custom_product_name:
-        for old_key, new_key in (
-            ("{family}", "{product[basetype]}"),
-            ("{asset}", "{folder[name]}"),
-            ("{subset}", "asset"),
-        ):
-            if old_key in custom_product_name:
-                log.warning(f"Using deprecated template key '{old_key}'")
-                custom_product_name = custom_product_name.replace(
-                    old_key, new_key
-                )
-
-        formatting_data = {
-            "folder": {
-                "name": folder_entity["name"]
-            },
-            "product": {
-                "name": product_name,
-                "type": product_type,
-                "basetype": product_base_type,
-            },
-        }
-        namespace = get_custom_namespace(
-            custom_product_name.format(**formatting_data)
-        )
+    namespace = get_rig_animation_instance_variant(context, namespace, options=options)
 
     if log:
         log.info("Creating product: {}".format(namespace))
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -122,8 +122,8 @@ def unlocked(node):
     """
     has_locked = cmds.lockNode(node, query=True, lock=True)[0]
     node_uuid = cmds.ls(node, uuid=True)[0]
+    cmds.lockNode(node, lock=False)
     try:
-        cmds.lockNode(node, lock=False)
         yield
 
     finally:
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -128,8 +128,11 @@ def unlocked(node):
 
     finally:
         if not cmds.objExists(node):
-            # Try and find the node by uuid
-            node = cmds.ls(node_uuid)[0]
+            # Node might have been renamed or deleted; try and find it by uuid.
+            nodes_from_id = cmds.ls(node_uuid, long=True) or []
+            if not nodes_from_id:
+                return
+            node = nodes_from_id[0]
         cmds.lockNode(node, lock=has_locked)
 
 
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -127,7 +127,8 @@ def unlocked(node):
         yield
 
     finally:
-        cmds.lockNode(node, lock=has_locked)
+        if cmds.objExists(node):
+            cmds.lockNode(node, lock=has_locked)
 
 
 @contextlib.contextmanager
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -4474,10 +4474,10 @@ def create_rig_animation_instance(
     )
     assert roots, "No root nodes in rig, this is a bug."
 
-    namespace = get_rig_animation_instance_variant(context, namespace, options=options)
+    variant = get_rig_animation_instance_variant(context, namespace, options=options)
 
     if log:
-        log.info("Creating product: {}".format(namespace))
+        log.info("Creating product: {}".format(variant))
 
     # Fill creator identifier
     creator_identifier = "io.openpype.creators.maya.animation"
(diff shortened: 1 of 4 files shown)


LiborBatek (approved):
In essence it works for both circumstances...when the settings `Update Namespace on Switch` been enabled and when disabled (legacy behaviour).

Upon inspection all seems working and well.

My only complaint is that when inspecting the Reference Editor one can see some leftovers after switching action...not neccessarilly a bad thing but there are some unwanted naming of the asset before switch (I went from `container` product and switched to `power generator` )

Image

Here is the namespace when freshly loaded the second asset `power generator` and its namespace been clean.

Image

Overall I think its acceptable and much better than before as in the Outliner / Scene Inventory everything looks correct!

I will leave it to you @moonyuet and @BigRoy  whether we should care about it or not really...

BigRoy:
> My only complaint is that when inspecting the Reference Editor one can see some leftovers after switching action...not neccessarilly a bad thing but there are some unwanted naming of the asset before switch (I went from `container` product and switched to `power generator` )

These names are from the reference nodes themselves. Interestingly enough I feel like these should be the EASIEST things to rename because they are really not tracked much - but by default you can't because they are locked nodes. (So I guess they are important to maya).

image

```python
from ayon_maya.api.lib import unlocked
from maya import cmds

with unlocked(ref_node):
    cmds.rename(ref_node, new_name)
```

This would technically work if the unlocked contextmanager internally wouldn't reference the node by name in the `finally:` block.
If we were to do `cmds.ls(nodes, uuid=True)` and track those instead we might get lucky (or somehow track them by what remains consistent in maya)

```python
node_uuid = cmds.ls(node, uuid=True)[0]
try:
    cmds.lockNode(node, lock=False)
finally:
    if not cmds.objExists(node):
        # Try and find the node by uuid
        node = cmds.ls(node_uuid)[0]
    cmds.lockNode(node, lock=has_locked)
```

Might be nice to also rename actually if we can @moonyuet .

client/ayon_maya/api/lib.py
```diff
@@ -127,7 +127,8 @@ def unlocked(node):
         yield
 
     finally:
-        cmds.lockNode(node, lock=has_locked)
+        if cmds.objExists(node):
+            cmds.lockNode(node, lock=has_locked)
```
BigRoy: This will now not lock the ref node - we should ensure it gets locked again. My pseudocode [here](https://github.com/ynput/ayon-maya/pull/457#issuecomment-5105589907) should allow to fix that. If you're afraid of adjusting the `unlocked` node we could perhaps pass a secondary argument `uuid=True` to enable to the UUID behavior, or use a separate context manager. The argument way might be the simplest.

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -393,6 +403,101 @@ def _set_display_handle(self, group_name: str):
         cmds.setAttr(f"{group_name}.selectHandleY", cy)
         cmds.setAttr(f"{group_name}.selectHandleZ", cz)
 
+    def _update_namespace(
+            self,
+            container: dict,
+            context: dict,
+            settings: dict,
```
BigRoy: Might be clearer as `loader_settings` to indicate it's not the full project settings?

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -393,6 +403,101 @@ def _set_display_handle(self, group_name: str):
...
+        namespace_template = settings.get("namespace")
+        if not namespace_template:
+            self.log.warning(
+                "No namespace template found in reference_loader settings; "
+                "skipping namespace update."
+            )
+            return
+        custom_namespace = namespace_template.format(
+            **self.get_namespace_template_data(context)
+        )
```
BigRoy: Could use `format_map` here I think?

```suggestion
        custom_namespace = namespace_template.format_map(
            self.get_namespace_template_data(context)
        )
```

Would be slightly more performant.

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -393,6 +403,101 @@ def _set_display_handle(self, group_name: str):
...
+            if members:
+                continue
+            # Ignore referenced object sets
+            if cmds.referenceQuery(object_set, isNodeReferenced=True):
+                continue
+            # Then only here confirm whether this is an animation instance, if so
+            # then we will want to auto-remove the instance
+            if get_creator_identifier(object_set) == "io.openpype.creators.maya.animation":
+                new_objectset = object_set.replace(old_namespace, new_namespace)
+                new_objectset = cmds.rename(object_set, new_objectset)
```
BigRoy: Where do we expect the namespace in the old animation instance name? I'm not too keen on just doing a `str.replace` because old namespace may be `cat` to change to `dog` and suddenly `cat_category:hello` would turn into `dog_dogegory:hello`.

Can we be more explicit?

Also, the object set != the product name. Also not too keen on just `str.replace()` on the product name - it's just too prone to edge cases. Same goes here, can we somehow be more explicit about the new name? Like somehow rely on just having it rename the animation instance to begin with as if it's a new instance?

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -393,6 +403,101 @@ def _set_display_handle(self, group_name: str):
...
+        for member in members:
+            object_sets.update(
+                cmds.listSets(object=member, extendToShape=False) or []
+            )
+        object_sets = cmds.ls(object_sets, type="objectSet")
+        for object_set in object_sets:
+            # Only consider empty object sets
+            members = cmds.sets(object_set, query=True)
+            if members:
+                continue
```
BigRoy: I don't understand this - why do we only consider empty ones? 🤔

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -257,6 +261,12 @@ def process_reference(self, context, name, namespace, options):
 
     def switch(self, container, context):
         self.update(container, context)
+        project_name = context["project"]["name"]
+        settings = get_project_settings(project_name)
```
BigRoy: Looking at the parent class there is `load_settings` attribute. Can we rely on it here too via `self.load_settings`?

server/settings/loaders.py
```diff
@@ -135,6 +135,9 @@ class ReferenceLoaderModel(BaseSettingsModel):
             "directly."
         )
     )
+    update_namespace_on_switch: bool = SettingsField(
+        title="Update Namespace on Switch"
```
BigRoy: This may need a more thorough tooltip/description, because it does more. It updates the namespace, for loaded rigs it also updates the animation instance product names, and it updates the reference node's name.

moonyuet: [219763d](https://github.com/ynput/ayon-maya/pull/457/commits/219763d9081b6686aad6b408d64831b8df033836)

client/ayon_maya/api/lib.py
moonyuet: [9dc0faf](https://github.com/ynput/ayon-maya/pull/457/commits/9dc0faf37a0ab4fa655cfb8e22bbbe0fbfdb1439)

client/ayon_maya/plugins/load/load_reference.py
moonyuet: [01d3706](https://github.com/ynput/ayon-maya/pull/457/commits/01d3706955b5d0e5528ef0bfa2f973d3546e98c0)

moonyuet: But the fix also means we need to test with remove rig

client/ayon_maya/plugins/load/load_reference.py
moonyuet: [9fc1b5b](https://github.com/ynput/ayon-maya/pull/457/commits/9fc1b5b1cff2549d1697df684d5e1de8d71cd2b9)

client/ayon_maya/plugins/load/load_reference.py
moonyuet: [9fc1b5b](https://github.com/ynput/ayon-maya/pull/457/commits/9fc1b5b1cff2549d1697df684d5e1de8d71cd2b9)

client/ayon_maya/plugins/load/load_reference.py
moonyuet: [fbeb803](https://github.com/ynput/ayon-maya/pull/457/commits/fbeb803beccfefa9d0a16cddb9d21e1f4b214ff5)

client/ayon_maya/api/lib.py
```diff
@@ -121,12 +121,18 @@ def unlocked(node):
         node (str): The name of the node to unlock.
     """
     has_locked = cmds.lockNode(node, query=True, lock=True)[0]
-    cmds.lockNode(node, lock=False)
-
+    node_uuid = cmds.ls(node, uuid=True)[0]
     try:
+        cmds.lockNode(node, lock=False)
```
BigRoy: I think lock node can remain out of the `try:` like before.
```suggestion
    cmds.lockNode(node, lock=False)
    try:
```

client/ayon_maya/api/plugin.py
```diff
@@ -796,6 +779,48 @@ def get_custom_namespace_and_group(self, context, options, loader_key):
...
+            context (dict): The context dictionary.
+            loader_key (str): The loader key.
+
+        Returns:
+            str: The resolved namespace template.
+        """
+        settings = self.load_settings[loader_key]
+        namespace_template = settings.get("namespace")
+        formatting_data = self.get_namespace_template_data(context)
+        return namespace_template.format(**formatting_data)
```
BigRoy: Is this used anywhere? Can be removed?

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -393,6 +401,119 @@ def _set_display_handle(self, group_name: str):
...
+
+        # Recreate using canonical creator logic so naming + productName are correct
+        try:
+            create_rig_animation_instance(
+                nodes=members,
+                context=context,
+                namespace=new_namespace,
+                options={"lock_instance": lock_instance},
+                log=self.log
+            )
```
BigRoy: Unfortunately I don't think this is the easiest approach either.

The created rig animation instance will not preserve any attribute overrides the original ones, it might suddenly target a different context, etc.

Preferably, we'd be able to reliably get the namespace/variant etc a new instance would receive on new creation, and then update the existing instance by overriding only those things.

We'd really be updating `variant` (I think the namespace is part of the variant?) and `productName`.
I think it'll be much more work to get a newly created animation instance to still work reliably with everything we may want to preserve in the scene file.

So we'd take out the variant/namespace logic from `create_rig_animation_instance` and make that a separate function we can call?

So we'd do:

```
variant = get_rig_animation_instance_variant(context, new_namespace)
product_name = CreateAnimation().get_product_name(
    project_name=project_name,
    folder_entity=folder_entity,  # folder should be the folder from the source instance
    task_entity=task_entity,  # task should be the task from the source instance
    variant=variant,
    instance=None,
    ... # etc
)
```
And take it from there, right?

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -393,6 +401,119 @@ def _set_display_handle(self, group_name: str):
...
+            node=container_node,
+            attribute="namespace",
+            value=new_namespace
+        )
+
+        cmds.namespace(rename=(old_namespace, new_namespace))
+
+        with unlocked(reference_node):
+            new_ref_node = reference_node.replace(old_namespace, new_namespace)
+            cmds.rename(reference_node, new_ref_node)
```
BigRoy: Can we define the new ref node name without `str.replace()` but compute the reference node name as if we were loading it new?

client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -393,6 +401,119 @@ def _set_display_handle(self, group_name: str):
...
+
+        container["namespace"] = new_namespace
```
BigRoy: Is this needed? 🤔 Don't think this `container` object still ends up being used anywhere, right?

moonyuet: I have added comment on that [a525372](https://github.com/ynput/ayon-maya/pull/457/commits/a52537252957c723737a02990c965181a75769da)

client/ayon_maya/plugins/load/load_reference.py
moonyuet: See: [a525372](https://github.com/ynput/ayon-maya/pull/457/commits/a52537252957c723737a02990c965181a75769da) But it is a bit hard-coded

client/ayon_maya/plugins/load/load_reference.py
moonyuet: I implemented this [f36ba9e](https://github.com/ynput/ayon-maya/pull/457/commits/f36ba9e7449775b72187d70f17744c152bd98e7d)

client/ayon_maya/api/plugin.py
moonyuet: [1717654](https://github.com/ynput/ayon-maya/pull/457/commits/17176549921c14c96c03763ae436b81ad81ab6a9)

client/ayon_maya/api/lib.py
```diff
@@ -4383,40 +4474,7 @@ def create_rig_animation_instance(
...
-            "product": {
-                "name": product_name,
-                "type": product_type,
-                "basetype": product_base_type,
-            },
-        }
-        namespace = get_custom_namespace(
-            custom_product_name.format(**formatting_data)
-        )
+    namespace = get_rig_animation_instance_variant(context, namespace, options=options)
```
BigRoy: Can we rename this variable `variant =` to avoid confusion from here on out.

client/ayon_maya/plugins/load/load_reference.py
BigRoy: But container metadata where? For the loader UI? Isn't the loader UI refreshing after an update? It's not using the actual `container` object that is passed into `update()` method right?

This `container` object is not returned - so if we're changing this key-value, where is that being picked up after this? isn't this variable garbage collected at that point?

client/ayon_maya/plugins/load/load_reference.py
BigRoy: I'm still confused here - why would we care only about empty object sets to begin with? wouldn't we care only about whether it's an animation instance or not?

BigRoy (commented):
Nice - just one more question and a small cosmetic change.

I'd say it's ready for testing by @LiborBatek now.

client/ayon_maya/plugins/load/load_reference.py
moonyuet: [be3a1f0](https://github.com/ynput/ayon-maya/pull/457/commits/be3a1f0d5a00536a38aa364a5608f43a9ec8dc86)

client/ayon_maya/plugins/load/load_reference.py
moonyuet: [bf745c2](https://github.com/ynput/ayon-maya/pull/457/commits/bf745c20ec9ec146c3923c91e20a997bcb3fae70)

client/ayon_maya/api/lib.py
moonyuet: [90e037f](https://github.com/ynput/ayon-maya/pull/457/commits/90e037f15c156e2b1446a8742307735e3896c235)

LiborBatek (approved):
I have tested with multiple product types and in `maya 2026` and `maya 2027` and all been working nicely atm!

Here all names align well...no traces of preceding/former product...

Image

PR #459, opened 2026-08-03
Bug: handle deepexr extension

### LINKED ISSUE #458: YN-0984: wrong expected files with deep exr in maya

Trying to publish rendering of Deep EXRs from Maya will fail, because the expected files have wrong extension - `Main.1001.deepexr`. 

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0984]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=fe547cb3e96d4815abbc5ead397b4735)

antirotor:
## Changelog Description
When setting the renderer to deep exr, AYON detected the file extension as `deepexr`. This fix is changing it to `exr.

*imfPluginKey* for deepexr is `deepexr` - correct it to just `exr` in that case.

Closed #458

## Testing notes:
Try to render (and publish) deep exr sequence in Maya.

### DIFF
--- a/client/ayon_maya/api/lib_renderproducts.py
+++ b/client/ayon_maya/api/lib_renderproducts.py
@@ -522,7 +522,7 @@ class RenderProductsArnold(ARenderProducts):
         "png": "png",
         "tiff": "tif",
         "mtoa_shaders": "ass",  # TODO: research what those last two should be
-        "maya": "",
+        "maya": "maya",
     }
 
     def get_renderer_prefix(self):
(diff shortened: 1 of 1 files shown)


moonyuet (approved):

client/ayon_maya/api/lib_renderproducts.py
```diff
@@ -714,6 +714,9 @@ def get_render_products(self):
         ]
 
         default_ext = self._get_attr("defaultRenderGlobals.imfPluginKey")
+        # Normalize imfPluginKey to expected file extension (deepexr -> exr)
+        if default_ext == "deepexr":
+            default_ext = self.aiDriverExtension["deepexr"]
```
BigRoy: Shouldn't [this mapping](https://github.com/ynput/ayon-maya/blob/bugfix%2Fyn-0984-wrong-expected-files-with-deep-exr-in-maya/client/ayon_maya/api/lib_renderproducts.py#L521) be used for this?

BigRoy: Took a quick look, this also applies to `jpeg` so we should use the mapping. The others, like `maya` or `tiff` actually come out correctly from `imfPluginKey` as `tif` and `maya` (which matches the resulting extension).

Either we add a dedicated `img_plugin_key_to_ext` mapping that contains:

```
img_plugin_key_to_ext = {
    "deepexr": "exr",
    "jpeg": "jpg",
}
```
Or we use the existing mapping (even though that contains more entries that actually don't end up being used.

After that, happy to merge.

BigRoy (approved):

PR #460, opened 2026-08-03
fix: temporarily disable the smooth warning in lib.displaySmoothness

vincentullmann:
## Changelog Description
temporarily disable the smooth warning popup when exporting Models

in `ExtractModel` and `ExtractObj` we change the display smoothness to 1 and then revert back to the original value.
If the model has a high poly count we may trigger a popup warning, which can stop the publish process

image

attribute ref: https://download.autodesk.com/us/maya/2010help/Nodes/mesh.html#attrsmoothWarn
image

## Testing notes (before this PR):
1. open maya
2. create a plane. set its subdivisions to 500x500 (or more)
3. click publish
4. go grab a coffee while you wait for the publish to process
5. come back to your desk to notice that the model has not been published, but instead you got this lovely popup

## Testing notes (with this PR):
1. repeat steps 1-4 from above
5. publish completes without any maya popups

## Testing notes (in python form):
```py
from maya import cmds
from ayon_maya.api import lib

# setup test geometry
my_plane, shape = cmds.polyPlane(sx=500, sy=500)  # this will show an warning
cmds.displaySmoothness(polygonObject=3)           # so will this

# this will show a warning
cmds.displaySmoothness(polygonObject=1)
cmds.displaySmoothness(polygonObject=3)

# no warning here
with lib.displaySmoothness(
    my_plane,
    polygonObject=1,
):
    print("Hello, world!")
    

# this will still show a warning
cmds.displaySmoothness(polygonObject=1)
cmds.displaySmoothness(polygonObject=3)
```

### DIFF
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -1153,21 +1153,28 @@ def displaySmoothness(nodes,
 
     originals = dict((node, parse(node)) for node in nodes)
 
-    try:
-        # Apply current state
-        cmds.displaySmoothness(nodes,
-                               divisionsU=divisionsU,
-                               divisionsV=divisionsV,
-                               pointsWire=pointsWire,
-                               pointsShaded=pointsShaded,
-                               polygonObject=polygonObject)
-        yield
-    finally:
-        # Revert state
-        _iteritems = getattr(originals, "iteritems", originals.items)
-        for node, state in _iteritems():
-            if state:
-                cmds.displaySmoothness(node, **state)
+    # Temporarily disable the smooth warning dialog for meshes
+    meshes = cmds.ls(nodes, type="mesh") or []
+    disable_smooth_warn = {f"{mesh}.smoothWarn": False for mesh in meshes}
+
+    with attribute_values(disable_smooth_warn):
+        try:
+            # Apply current state
+            cmds.displaySmoothness(
+                nodes,
+                divisionsU=divisionsU,
+                divisionsV=divisionsV,
+                pointsWire=pointsWire,
+                pointsShaded=pointsShaded,
+                polygonObject=polygonObject,
+            )
+            yield
+        finally:
+            # Revert state
+            _iteritems = getattr(originals, "iteritems", originals.items)
+            for node, state in _iteritems():
+                if state:
+                    cmds.displaySmoothness(node, **state)
 
 
 @contextlib.contextmanager
--- a/client/ayon_maya/api/lib.py
+++ b/client/ayon_maya/api/lib.py
@@ -1171,8 +1171,7 @@ def displaySmoothness(nodes,
             yield
         finally:
             # Revert state
-            _iteritems = getattr(originals, "iteritems", originals.items)
-            for node, state in _iteritems():
+            for node, state in originals.items():
                 if state:
                     cmds.displaySmoothness(node, **state)
 


client/ayon_maya/api/lib.py
```diff
@@ -1166,9 +1166,18 @@ def parse(node):
...
+            smooth_warn_attr = f"{node}.smoothWarn"
+            smooth_warn = cmds.getAttr(smooth_warn_attr)
+            cmds.setAttr(smooth_warn_attr, False)
+
             if state:
                 cmds.displaySmoothness(node, **state)
 
+            # restore smooth warn state
+            cmds.setAttr(smooth_warn_attr, smooth_warn)
+
```
moonyuet: I think we can get the attribute of smoothWarn before yielding first and then we can reuse it to reset the smoothWarn back to the original value. And this would ensure we can always set the value back to original state when setting it to False during context.

vincentullmann: yeah, good point.

initially I was trying to minimize the time the `smoothWarn` is turned off, which is why I had it set only aroud the lines that restore the original state.

I've changed it now to be set on entering the context, which means we are now able to do this, without getting an popup
```py
with lib.displaySmoothness(my_object, polygonObject=3):
    pass
```

client/ayon_maya/api/lib.py
```diff
@@ -1151,9 +1151,20 @@ def parse(node):
...
 
-    originals = dict((node, parse(node)) for node in nodes)
+    # store original values
+    original_values = {
+        node: {
+            "displaySmoothness": parse(node),
+            "smoothWarn": cmds.getAttr(f"{node}.smoothWarn"),
+        }
+        for node in nodes
+    }
```
BigRoy: This will fail on non-mesh objects, like nurbs surfaces and curves because they do not have the `smoothWarn` attribute. Please make sure the object has `smoothWarn` attribute or that the object is a mesh.

This function allows to be run on other objects than polygonal ones, so we should make sure it works fine for them too.

BigRoy (changes requested):

client/ayon_maya/api/lib.py
vincentullmann: good point.
I've updated the PR to only apply to meshes.

tested with a mesh, nurbs and a light:
```py
nodes = ["pPlane1", "nurbsCylinder1", "directionalLight1"]
with lib.displaySmoothness(nodes, polygonObject=3):
    pass
 ```

File: client/ayon_maya/api/lib.py [hunk 3]
```diff
@@ -1151,9 +1152,20 @@ def parse(node):
...
 
-    originals = dict((node, parse(node)) for node in nodes)
+    # store original values
+    original_values = {}
+    for node in nodes:
+        original_values[node] = {"displaySmoothness": parse(node)}
+
+    # for meshes also store the smoothWarn
+    for mesh in meshes:
+        original_values[mesh]["smoothWarn"] = cmds.getAttr(f"{mesh}.smoothWarn")
```
BigRoy: Have you considered just doing this:

```python
# Disable smooth warning pop-up on meshes
meshes = cmds.ls(nodes, type="mesh", long=True)
disable_smooth_warn = {
    f"{mesh}.smoothWarn": False 
    for mesh in cmds.ls(nodes, types="mesh", long=True)
}
with attribute_values(disable_smooth_warn):
    ...
```
It'd require no data model changes, so you don't need the dict and you can restore lots of the old logic which may feel quite a bit simpler? 🤔

vincentullmann: yeah, 

I was looking for a context manager like `attribute_values` but didn't find it. Should have looked closer it seems :D

client/ayon_maya/api/lib.py
```diff
@@ -1153,21 +1153,28 @@ def parse(node):
...
+                divisionsV=divisionsV,
+                pointsWire=pointsWire,
+                pointsShaded=pointsShaded,
+                polygonObject=polygonObject,
+            )
+            yield
+        finally:
+            # Revert state
+            _iteritems = getattr(originals, "iteritems", originals.items)
+            for node, state in _iteritems():
```
BigRoy: `iteritems` is from Python 2-era.
```suggestion
            for node, state in originals.items():
```

BigRoy (approved):
LGTM - one more comment, then needs some testing and ready to go. @moonyuet can you test it?

moonyuet (approved):
Tested successfully in Maya 2027 

[publish-report-260817-13-00.json](https://github.com/user-attachments/files/31129666/publish-report-260817-13-00.json)

PR #464, opened 2026-08-21
MayaUSD: Fix the duplicate overrides in USD options

### LINKED ISSUE #461: Duplicate Publisher GUI Elements when Exposed Overrides

## Current Behavior
When I have selected all the items in the settings enumerator `ayon+settings://maya/publish/ExtractMayaUsd/overrides`, and then I create a Maya USD Model, I get double attributes in the GUI of the Ayon Publisher when togglng on the Extract USD:

## Expected Behavior
Not having double elements.

Image

Image

### Steps To Reproduce:

Set the `ayon+settings://maya/publish/ExtractMayaUsd/overrides` to select all.

1. Open Maya
2. Create Maya USD Model
3. Toggle on Extract USD
4. Look at the `USD Options` attributes, they are duplicated.

### Version
ayon-server 1.16.0
ayon-core 1.9.8
ayon-maya 0.6.10

### What platform were you running when you found the bug?
Windows

### Relevant log output:

```shell

```

moonyuet:
## Changelog Description
This PR is to remove the duplicate overrides in USD options in MayaUSD publish. This also avoids the two extractors exporting the assets twices which error out the integrator.
Resolve https://github.com/ynput/ayon-maya/issues/461

Marked it as draft as it needs more clarification for the legacy creator.

## Testing notes:
1. Create Maya USD instance with your model hierarchies.
2. Publish

### DIFF
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -259,4 +259,4 @@ class CreateMayaUsdModel(CreateMayaUsd):
         return attr_defs
 
     def get_publish_families(self):
-        return ["model", "usd"]
+        return ["model", "usd", "mayaUsd.filterProperties"]
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -239,7 +239,7 @@ class CreateMayaUsd(plugin.MayaCreator):
 
 
 class CreateMayaUsdModel(CreateMayaUsd):
-    identifier = "io.ayon.creators.maya.mayamayaUsd.model"
+    identifier = "io.ayon.creators.maya.mayaUsd.model"
     label = "Maya USD: Model"
     product_base_type = "model"
     product_type = product_base_type
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -257,3 +257,6 @@ class CreateMayaUsdModel(CreateMayaUsd):
                 attr_def.default = True
 
         return attr_defs
+
+    def get_publish_families(self):
+        return ["model"]
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -239,7 +239,7 @@ class CreateMayaUsd(plugin.MayaCreator):
 
 
 class CreateMayaUsdModel(CreateMayaUsd):
-    identifier = "io.ayon.creators.maya.mayamayaUsd.model"
+    identifier = "io.ayon.creators.maya.mayausd.model"
     label = "Maya USD: Model"
     product_base_type = "model"
     product_type = product_base_type
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -259,4 +259,4 @@ class CreateMayaUsdModel(CreateMayaUsd):
         return attr_defs
 
     def get_publish_families(self):
-        return ["model", "usd", "mayaUsd.filterProperties"]
+        return ["usd.model", "usd", "mayaUsd.filterProperties"]
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -259,4 +259,4 @@ class CreateMayaUsdModel(CreateMayaUsd):
         return attr_defs
 
     def get_publish_families(self):
-        return ["model"]
+        return ["usd", "model"]
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -239,7 +239,7 @@ class CreateMayaUsd(plugin.MayaCreator):
 
 
 class CreateMayaUsdModel(CreateMayaUsd):
-    identifier = "io.ayon.creators.maya.mayausd.model"
+    identifier = "io.ayon.creators.maya.mayamayaUsd.model"
     label = "Maya USD: Model"
     product_base_type = "model"
     product_type = product_base_type
@@ -259,4 +259,4 @@ class CreateMayaUsdModel(CreateMayaUsd):
         return attr_defs
 
     def get_publish_families(self):
-        return ["usd.model", "usd", "mayaUsd.filterProperties"]
+        return ["usd", "mayaUsd.model", "mayaUsd.filterProperties"]
--- a/client/ayon_maya/plugins/create/create_maya_usd.py
+++ b/client/ayon_maya/plugins/create/create_maya_usd.py
@@ -259,4 +259,4 @@ class CreateMayaUsdModel(CreateMayaUsd):
         return attr_defs
 
     def get_publish_families(self):
-        return ["usd", "model"]
+        return ["model", "usd"]
--- a/client/ayon_maya/plugins/create/create_model.py
+++ b/client/ayon_maya/plugins/create/create_model.py
@@ -18,6 +18,9 @@ class CreateModel(plugin.MayaCreator):
     write_face_sets = True
     include_shaders = False
 
+    def get_publish_families(self):
+        return ["model", "model.extract"]
+
     def get_instance_attr_defs(self):
 
         return [
--- a/client/ayon_maya/plugins/create/create_model.py
+++ b/client/ayon_maya/plugins/create/create_model.py
@@ -19,7 +19,7 @@ class CreateModel(plugin.MayaCreator):
     include_shaders = False
 
     def get_publish_families(self):
-        return ["model", "model.extract"]
+        return ["model.extract"]
 
     def get_instance_attr_defs(self):
 
--- a/client/ayon_maya/plugins/publish/extract_maya_usd.py
+++ b/client/ayon_maya/plugins/publish/extract_maya_usd.py
@@ -532,7 +532,7 @@ class ExtractMayaUsd(plugin.MayaExtractorPlugin,
                 continue
 
             publish_attributes = value_changes["publish_attributes"]
-            class_name = cls.__name__
+            class_name = cls.__class__.__name__
             if class_name not in publish_attributes:
                 continue
 
@@ -558,7 +558,7 @@ class ExtractMayaUsd(plugin.MayaExtractorPlugin,
             plugin_attr_values = (
                 instance.data
                 .get("publish_attributes", {})
-                .get(cls.__name__, {})
+                .get(cls.__class__.__name__, {})
             )
             is_enabled = plugin_attr_values.get("active", cls.active)
         defs = super().get_attr_defs_for_instance(create_context, instance)
@@ -866,6 +866,13 @@ class ExtractMayaUsdModel(ExtractMayaUsd):
 
     def process(self, instance):
         # TODO: Fix this without changing instance data
+        families = set(instance.data.get("families") or [])
+        if "mayaUsd" in families:
+            self.log.debug(
+                "Skipping ExtractMayaUsdModel because instance already "
+                "matches dedicated mayaUsd extractor."
+            )
+            return
         instance.data["exportAnimationData"] = False
         super().process(instance)
 
--- a/client/ayon_maya/plugins/publish/extract_maya_usd.py
+++ b/client/ayon_maya/plugins/publish/extract_maya_usd.py
@@ -837,13 +837,6 @@ class ExtractMayaUsdModel(ExtractMayaUsd):
 
     def process(self, instance):
         # TODO: Fix this without changing instance data
-        families = set(instance.data.get("families") or [])
-        if "mayaUsd" in families:
-            self.log.debug(
-                "Skipping ExtractMayaUsdModel because instance already "
-                "matches dedicated mayaUsd extractor."
-            )
-            return
         instance.data["exportAnimationData"] = False
         super().process(instance)
 
--- a/client/ayon_maya/plugins/publish/extract_maya_usd.py
+++ b/client/ayon_maya/plugins/publish/extract_maya_usd.py
@@ -858,7 +858,7 @@ class ExtractMayaUsdModel(ExtractMayaUsd):
     """
 
     label = "Extract USD"
-    families = ["model", "mayaUsd.model"]
+    families = ["mayaUsd.model"]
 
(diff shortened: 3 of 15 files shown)


client/ayon_maya/plugins/publish/extract_maya_usd.py
```diff
@@ -866,6 +866,13 @@ class ExtractMayaUsdModel(ExtractMayaUsd):
 
     def process(self, instance):
         # TODO: Fix this without changing instance data
+        families = set(instance.data.get("families") or [])
```
moonyuet: @BigRoy I have a strong feeling we shouldn't do this but instead we should implement the `get_published_families()` to return just `["model"]`  in modelExtractor.

client/ayon_maya/plugins/publish/extract_maya_usd.py
```diff
@@ -532,7 +532,7 @@ def on_values_changed(cls, event):
                 continue
 
             publish_attributes = value_changes["publish_attributes"]
-            class_name = cls.__name__
+            class_name = cls.__class__.__name__
```
BigRoy: Wait... this is wrong though? Isn't the class of a class `type`?

`cls` is already a class.

client/ayon_maya/plugins/publish/extract_maya_usd.py
BigRoy: Yup, we should preferably make it so it just doesn't trigger both to begin with, so we'll need to be more explicit about the families I think. 🤔

client/ayon_maya/plugins/publish/extract_maya_usd.py
moonyuet: Yes it is. It is meant to be testing the duplicate overrides. I will push the right commits.

moonyuet: [5c6f94d](https://github.com/ynput/ayon-maya/pull/464/commits/5c6f94d5b60ca72741f2b1761e0329b230c5435d)

client/ayon_maya/plugins/publish/extract_maya_usd.py
moonyuet: [9766be1](https://github.com/ynput/ayon-maya/pull/464/commits/9766be1e33501b41f252bae95e3e22288338f44c)

client/ayon_maya/plugins/publish/extract_maya_usd.py
moonyuet: [5c6f94d](https://github.com/ynput/ayon-maya/pull/464/commits/5c6f94d5b60ca72741f2b1761e0329b230c5435d)

client/ayon_maya/plugins/create/create_maya_usd.py
```diff
@@ -257,3 +257,6 @@ def get_pre_create_attr_defs(self):
                 attr_def.default = True
 
         return attr_defs
+
+    def get_publish_families(self):
+        return ["model"]
```
BigRoy: Doesn't this make it so that it **does not** have the USD contribution attributes? 🤔 If so, that's wrong. :)

moonyuet: [5cfd6c5](https://github.com/ynput/ayon-maya/pull/464/commits/5cfd6c5dcc6a27c5f5d0d013b227c70c4ff6e0f6),
Publish report:
[publish-report-260821-20-13.json](https://github.com/user-attachments/files/31305159/publish-report-260821-20-13.json)

client/ayon_maya/plugins/publish/extract_maya_usd.py
```diff
@@ -516,35 +516,6 @@ def parse_attr_str(attr_str):
...
-                continue
-
-            if "active" not in publish_attributes[class_name]:
-                continue
-
-            # Update the attribute definitions
-            new_attrs = cls.get_attr_defs_for_instance(
-                event["create_context"], instance
-            )
-            instance.set_publish_plugin_attr_defs(class_name, new_attrs)
```
BigRoy, line 546: Don't we still need this so that the attribute update/hide based on whether "Active" toggle is on or off?

client/ayon_maya/plugins/create/create_maya_usd.py
```diff
@@ -257,3 +257,6 @@ def get_pre_create_attr_defs(self):
                 attr_def.default = True
 
         return attr_defs
+
+    def get_publish_families(self):
+        return ["usd", "model"]
```
BigRoy: Just for piece of mind, this is more model than usd so I like to keep that first in the list to be clear about that.
```suggestion
    def get_publish_families(self):
        return ["model", "usd"]
```

client/ayon_maya/plugins/publish/extract_maya_usd.py
moonyuet: [1099cd2](https://github.com/ynput/ayon-maya/pull/464/commits/1099cd2e98ef79524e7b4f824deb7d5737797471)

BigRoy (commented):
Nice - this could work.

I think we'll need to add a dedicated additional family to target this plug-in: https://github.com/ynput/ayon-maya/blob/44ef2cef3eff818b6fd561fc9f948341a081d31b/client/ayon_maya/plugins/publish/collect_maya_usd_export_filter_properties.py#L8

So we can still have it apply for the `mayaUsd` ones but can also enable it for this one without it by default supporting `model` family, e.g. a `mayaUsd.filterProperties` family or alike?

moonyuet:
@BigRoy can you review?

BigRoy (approved):
Code LGTM - but didn't have time to test now. @LiborBatek any chance you can give it a go today?

- [ ] Maya USD family should still work
- [ ] Maya USD: Model product should work, display attributes correctly and integrate with USD file correctly.
    - I think we may have an issue here @moonyuet because I think this one will now also export the default `model` outputs which it previously didn't if I'm not mistaken. I believe previously it ONLY exported the USD. Can you confirm?
- [ ] Regular model product also still works, without USD

client/ayon_maya/plugins/create/create_maya_usd.py
```diff
@@ -257,3 +257,6 @@ def get_pre_create_attr_defs(self):
                 attr_def.default = True
 
         return attr_defs
+
+    def get_publish_families(self):
+        return ["model", "usd", "mayaUsd.filterProperties"]
```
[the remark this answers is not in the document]
moonyuet: [a2213be](https://github.com/ynput/ayon-maya/pull/464/commits/a2213bedb0e91c96699feb5cbe7388355fefae99)

moonyuet:
>   * I think we may have an issue here @moonyuet because I think this one will now also export the default `model` outputs which it previously didn't if I'm not mistaken. I believe previously it ONLY exported the USD. Can you confirm?

I noticed that just now. And already implement the explicit publish families for usd.model

BigRoy (changes requested):
The fix is getting better and better, but there's still an annoying edge case. The extractor being used now, applies to both `model` and this particular creator now. However, the default state of this extractor is to be `active = False` by default. However, for the `Maya USD: Model` we do not ever want it to really be optional by default.

Also it currently still seems to be listing the Model (Maya Scene) and Alembic options as well, which shouldn't be the case for this instance. It looks like the `model` family still applies (which I think technically is correct, because we will want the model validators to trigger).

Image

I think what we'll need to do is make the regular Model creator have an additional family that then the Extractors for `model` apply to, e.g. `model.extract` or whatever.

So we'd almost need a separate inherited Extract Model USD plug-in that is solely targeting this family but is enforced to always be enabled and not optional? And we'll also need to avoid the conflict with the regular "model" extractors.

That the model validators apply, is **correct** - but we don't want any of its extractors (and vice-versa).

client/ayon_maya/plugins/create/create_model.py
```diff
@@ -18,6 +18,9 @@ class CreateModel(plugin.MayaCreator):
     write_face_sets = True
     include_shaders = False
 
+    def get_publish_families(self):
+        return ["model", "model.extract"]
```
iLLiCiTiT:
```suggestion
        return ["model.extract"]
```

LiborBatek (approved):
I have tested through publishing of the `model` `mayaUSD:Model` and `MayaUSD:usdAsset` as model and all been publishing successfully.

I havent experienced any duplication of the USD overrides while having those enabled (aka exposed)...all good on this front.

LGTM

P.S. one small thing Im curious about is if when using standard `modelMain` and having enabled  `ayon+settings://maya/publish/ExtractMayaUsd` if the `USD representation` could be as `optional` as I couldnt achieve it and it was always produced too leading to e.g. `.ma` `.abc` and `.usd` allways... Im not saying its a flaw but been curious if it is correct or not. @BigRoy

BigRoy:
> P.S. one small thing Im curious about is if when using standard `modelMain` and having enabled `ayon+settings://maya/publish/ExtractMayaUsd` if the `USD representation` could be as `optional` as I couldnt achieve it and it was always produced too leading to e.g. `.ma` `.abc` and `.usd` allways... Im not saying its a flaw but been curious if it is correct or not. @BigRoy

Standard `model` product type should indeed allow this to be optional, if it isn't - it's a bug somewhere.

moonyuet:
> > P.S. one small thing Im curious about is if when using standard `modelMain` and having enabled `ayon+settings://maya/publish/ExtractMayaUsd` if the `USD representation` could be as `optional` as I couldnt achieve it and it was always produced too leading to e.g. `.ma` `.abc` and `.usd` allways... Im not saying its a flaw but been curious if it is correct or not. @BigRoy
> 
> Standard `model` product type should indeed allow this to be optional, if it isn't - it's a bug somewhere.

It could have been because the last commit inheriting unnecessary class. It should be cleaned up and I have added the `model.extract` into families of ExtractUSD.
[eafda4c](https://github.com/ynput/ayon-maya/pull/464/commits/eafda4c037e92d16f7fc908c9fcee4cdfb642991)

BigRoy:
> It could have been because the last commit inheriting unnecessary class. It should be cleaned up and I have added the `model.extract` into families of ExtractUSD.
> [eafda4c](https://github.com/ynput/ayon-maya/pull/464/commits/eafda4c037e92d16f7fc908c9fcee4cdfb642991)

However, with this change - then we don't get the enforced disabling of the `exportAnimationData` which I wonder if we should retain?

moonyuet:
> > It could have been because the last commit inheriting unnecessary class. It should be cleaned up and I have added the `model.extract` into families of ExtractUSD.
> > [eafda4c](https://github.com/ynput/ayon-maya/pull/464/commits/eafda4c037e92d16f7fc908c9fcee4cdfb642991)
> 
> However, with this change - then we don't get the enforced disabling of the `exportAnimationData` which I wonder if we should retain?

It is not retained anymore.
[04348da](https://github.com/ynput/ayon-maya/pull/464/commits/04348dae1c89049348dfba1abba3d854a4213cd9)

client/ayon_maya/plugins/publish/extract_maya_usd.py
```diff
@@ -858,16 +858,11 @@ class ExtractMayaUsdModel(ExtractMayaUsd):
...
+    families = ["mayaUsd.model", "model.extract"]
 
     # Exposed in settings
     optional = True
-    active = False
-
-    def process(self, instance):
-        # TODO: Fix this without changing instance data
-        instance.data["exportAnimationData"] = False
-        super().process(instance)
```
BigRoy, line 870: Uh wait a second - this is now removed **but** it **should** be applied to models right?

moonyuet: I removed due to https://github.com/ynput/ayon-maya/pull/464#issuecomment-5516065192 . I reverted it back now. Do you mean `model.extract` should be added into the `ExtractMayaUsd` families instead?

BigRoy: What I meant was that the model exports should still ensure `instance.data["exportAnimationData"] = False` - _likely_ for both of them, but I'm 100% sure that at least it should be the case for the 'regular' model product type creator. Whether it's also for the Maya Model one, I _think so_ but I'm not certain. depends on whether that exposes a frame range attribute to the user, if it doesn't (which I'd say is correct) then this override would likely need to be there.

moonyuet: Already reverted back: [677c402](https://github.com/ynput/ayon-maya/pull/464/commits/677c402a4074c1f481d0ffc55179b730da3ae4e9). I just reverted it back just to be safe. (If the attributes are not turned on by default, it won't be there. But there are users customizing their setup to test and they would edit manually).

client/ayon_maya/plugins/publish/extract_maya_usd.py
```diff
@@ -858,11 +858,11 @@ class ExtractMayaUsdModel(ExtractMayaUsd):
     """
 
     label = "Extract USD"
-    families = ["model"]
+    families = ["mayaUsd.model", "model.extract"]
```
BigRoy: This has the annoying side effect that it can now also be optional for `mayaUsd.model` but for that one, it should honestly never be - it's required for it.

moonyuet: I would say we can implement another class with label as `Extract USD for contribution` and with the optional value as False.

BigRoy: I think it's the only way, yes.

moonyuet: [1919155](https://github.com/ynput/ayon-maya/pull/464/commits/1919155d70cbb7960cfed1d69af74102eb761bec)

BigRoy (approved):
@LiborBatek can you validate this works for you?

I'll want to think a bit more about the `model.extract` naming - but admittedly, it's not required for getting this merged. So, approving. If this works, I'm all good.

BigRoy:
@moonyuet I've noticed that for `io.ayon.creators.maya.mayausd.model` creator the instance, also shows "Alembic options" near the bottom, but if I understand correctly - it shouldn't; right? 🤔 

image

moonyuet:
> @moonyuet I've noticed that for io.ayon.creators.maya.mayausd.model creator the instance, also shows "Alembic options" near the bottom, but if I understand correctly - it shouldn't; right? 🤔

[8a6b5d2](https://github.com/ynput/ayon-maya/pull/464/commits/8a6b5d2b5661d265a2363ead15be91a65802d3e5)

moonyuet:
@BigRoy @LiborBatek can you review again?

PR #465, opened 2026-08-21
Review: greasePencil and bluePencils into viewport options

### LINKED ISSUE #420: Extract Playblast supports Blue Pencil formerly named Grease Pencil not working even when correctly set

### Description

Autodesk Maya: Blue Pencil Support in ExtractPlayblast
AYON's ExtractPlayblast does not capture Blue Pencil annotations in review publishes, even when correctly configured in viewport settings.

Current Behavior
Blue Pencil annotations are visible in Maya viewport

Manual playblasts (via Maya UI) include Blue Pencil

AYON review publishes omit Blue Pencil annotations

Setting “Grease Pencils”: true in showhas no effect

Setting "bluePencil": true in pluginObjects has no effect

Root Cause
Maya uses separate optionVars for playblast display settings that are independent from modelEditor flags. The critical missing optionVars are:

playblastShowBluePencil

playblastOverrideViewport

### Steps To Reproduce:

1. Use newer maya with 'Blue Pencil' feature present e.g Maya 2026
2. Draw some notes over viewport
3. Extract Playblast with enabled former setting "Grease Pencils" set to true
4. Annotations been missing in the resulting reviewable mp4

### Version
Latest Ayon maya version

### What platform were you running when you found the bug?
Windows

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

moonyuet:
## Changelog Description
This PR is to add greasePencil and bluePencils as default viewport options. This also add the bluePencils into addon settings.
Resolve https://github.com/ynput/ayon-maya/issues/420

If only bluePencil is enabled, it is only working when `ayon+settings://maya/publish/ExtractPlayblast/profiles/0/capture_preset/ViewportOptions/override_viewport_options` disabled.
If you want to enable `ayon+settings://maya/publish/ExtractPlayblast/profiles/0/capture_preset/ViewportOptions/override_viewport_options`  and make the blue pencil working, you need to enable `ayon+settings://maya/publish/ExtractPlayblast/profiles/0/capture_preset/ViewportOptions/headsUpDisplay` (See the screenshots) 
image

## Testing notes:
1. Create Review
2. Draw with greasePencil or bluePencils
3. Enable `ayon+settings://maya/publish/ExtractPlayblast/profiles/0/capture_preset/ViewportOptions/bluePencil`
4. Publish

### DIFF
--- a/client/ayon_maya/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_maya/plugins/publish/extract_thumbnail.py
@@ -53,11 +53,12 @@ class ExtractThumbnail(plugin.MayaExtractorPlugin,
             instance, camera, path,
             start=1, end=1,
             capture_preset=capture_preset)
+        # Force thumbnail capture to JPG if blue pencil is enabled
+        # as png captures alpha channels which maps out only the
+        # blue pencil's annotation and it doesn't show the composited
+        # version of the thumbnail.
+        # This is temporary until a better solution is implemented.
         if preset["viewport_options"].get("bluePencil"):
-            # Force thumbnail capture to JPG if blue pencil is enabled
-            # as png captures alpha channels which maps out only the
-            # blue pencil's annotation and it doesn't show the composited
-            # version of the thumbnail.
             self.log.debug(
                 "Blue Pencil is enabled; forcing thumbnail capture to JPG "
                 "so annotations are composited over the thumbnail image "
--- a/client/ayon_maya/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_maya/plugins/publish/extract_thumbnail.py
@@ -53,6 +53,9 @@ class ExtractThumbnail(plugin.MayaExtractorPlugin,
             instance, camera, path,
             start=1, end=1,
             capture_preset=capture_preset)
+        if preset["viewport_options"].get("bluePencil"):
+            self.log.debug("Blue pencil is enabled, setting compression to jpg")
+            preset["compression"] = "jpg"
 
         preset["camera_options"].update({
             "displayGateMask": False,
--- a/client/ayon_maya/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_maya/plugins/publish/extract_thumbnail.py
@@ -54,6 +54,10 @@ class ExtractThumbnail(plugin.MayaExtractorPlugin,
             start=1, end=1,
             capture_preset=capture_preset)
         if preset["viewport_options"].get("bluePencil"):
+            # Force thumbnail capture to JPG if blue pencil is enabled
+            # as png captures alpha channels which maps out only the
+            # blue pencil's annotation and it doesn't show the composited
+            # version of the thumbnail.
             self.log.debug(
                 "Blue Pencil is enabled; forcing thumbnail capture to JPG "
                 "so annotations are composited over the thumbnail image "
--- a/client/ayon_maya/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_maya/plugins/publish/extract_thumbnail.py
@@ -54,7 +54,11 @@ class ExtractThumbnail(plugin.MayaExtractorPlugin,
             start=1, end=1,
             capture_preset=capture_preset)
         if preset["viewport_options"].get("bluePencil"):
-            self.log.debug("Blue pencil is enabled, setting compression to jpg")
+            self.log.debug(
+                "Blue Pencil is enabled; forcing thumbnail capture to JPG "
+                "so annotations are composited over the thumbnail image "
+                "instead of exported as a transparent annotation layer."
+            )
             preset["compression"] = "jpg"
 
         preset["camera_options"].update({
--- a/client/ayon_maya/vendor/python/capture.py
+++ b/client/ayon_maya/vendor/python/capture.py
@@ -731,11 +731,10 @@ def _applied_viewport_options(options, panel):
 
     options = dict(ViewportOptions, **(options or {}))
     plugin_options = options.pop("pluginObjects", {})
-    if "bluePencil" in options and "headsUpDisplay" not in options:
-        logger.warning(
-            "Blue Pencil is not visible as Heads Up Display is disabled."
-            "Enable Heads Up Display to see Blue Pencil."
-        )
+    if "bluePencil" in options:
+        logger.warning("Heads Up Display is required for Blue Pencil to be "
+                       "visible. Enabling it automatically.")
+        options["headsUpDisplay"] = True
     # BUGFIX Maya 2020 some keys in viewport options dict may not be unicode
     #        This is a local OpenPype edit to capture.py for issue #4730
     # TODO: Remove when dropping Maya 2020 compatibility
--- a/client/ayon_maya/vendor/python/capture.py
+++ b/client/ayon_maya/vendor/python/capture.py
@@ -731,7 +731,11 @@ def _applied_viewport_options(options, panel):
 
     options = dict(ViewportOptions, **(options or {}))
     plugin_options = options.pop("pluginObjects", {})
-
+    if "bluePencil" in options and "headsUpDisplay" not in options:
+        logger.warning(
+            "Blue Pencil is not visible as Heads Up Display is disabled."
+            "Enable Heads Up Display to see Blue Pencil."
+        )
     # BUGFIX Maya 2020 some keys in viewport options dict may not be unicode
     #        This is a local OpenPype edit to capture.py for issue #4730
     # TODO: Remove when dropping Maya 2020 compatibility
--- a/client/ayon_maya/vendor/python/capture.py
+++ b/client/ayon_maya/vendor/python/capture.py
@@ -12,7 +12,7 @@ import logging
 from maya import cmds
 from maya import mel
 
-from qtpy import QtGui, QtWidgets
+from qtpy import QtGui, QtWidgets, QtCore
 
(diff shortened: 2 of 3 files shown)


moonyuet:
@BigRoy @antirotor @LiborBatek Should we enforce not using override_vireport_options when we use bluePencils?

BigRoy:
> @BigRoy @antirotor @LiborBatek Should we enforce not using override_vireport_options when we use bluePencils?

No. It must also work with override viewport enabled, if the flag is enabled in settings.

Viewport Options override disabled means that it takes your current view's state... And then none of the settings apply. So essentially, it just happens to work because you had it enabled in viewport before publishing?

moonyuet:
> > @BigRoy @antirotor @LiborBatek Should we enforce not using override_vireport_options when we use bluePencils?
> 
> No. It must also work with override viewport enabled, if the flag is enabled in settings.
> 
> Viewport Options override disabled means that it takes your current view's state... And then none of the settings apply. So essentially, it just happens to work because you had it enabled in viewport before publishing?

It is quite maya-thing, you need to make sure head up display set to true when enabling the blue pencil. It is either to have the validator to check, or log a warning for users, or raise exception when users turn on blue pencil while head up display is off. Or maybe enforce head up display set to true when blue pencil is enabled.
But I would prefer the log message for that: [5a62f93](https://github.com/ynput/ayon-maya/pull/465/commits/5a62f93ed10829c91e2f80e4565042e42348b638) or validator.

client/ayon_maya/vendor/python/capture.py
```diff
@@ -729,7 +731,11 @@ def _applied_viewport_options(options, panel):
 
     options = dict(ViewportOptions, **(options or {}))
     plugin_options = options.pop("pluginObjects", {})
-
+    if options.get("bluePencil") and not options.get("headsUpDisplay"):
```
BigRoy, line 766: I wonder if it'd make sense instead to stack in an `_empty_hud` context manager so that if HUD is disabled for blue pencil that at least we try to hide as much as possible from the HUD but force enable it, like:

```python
import contextlib
from maya import cmds

@contextlib.contextmanager
def empty_hud():
    """Force all HUD elements to be disabled during context."""
    view_manip = cmds.viewManip(query=True, v=True)
    huds = cmds.headsUpDisplay(listHeadsUpDisplays=True)
    optionvars = {
        "inViewEditorVisible",
        "inViewMessageEnable"
    }
    enabled_optionvars: list[str] = []
    enabled_huds: list[str] = []
    try:    
        cmds.viewManip(v=False)
        for ov in optionvars:
            if cmds.optionVar(query=ov):
                enabled_optionvars.append(ov)
        for hud in huds:
            if cmds.headsUpDisplay(hud, query=True, visible=True):
                cmds.headsUpDisplay(hud, edit=True, visible=False)
                enabled_huds.append(hud)
        yield
    finally:
        cmds.viewManip(v=view_manip)
        for hud in enabled_huds:
            cmds.headsUpDisplay(hud, edit=True, visible=True)
        for ov in enabled_optionvars:
            cmds.optionVar(intValue=(ov, 1))
```

So that instead of warning the user, we could potentially safely enable the HUD but exclude any HUD parts. 

**There's just the side effect that we may not know whether indeed we're hiding all cases where showing HUD may affect something else.** So I guess for now perhaps best is to just warn instead.

BigRoy (approved):
LGTM

LiborBatek (approved):
I have tested it with `maya 2027` and `blue pencil` tool it works but...

generally speaking it works when enabled `display overrides` which is correct and also `HUD` enabled...

but it only works when those enabled so the `Grease Pencil` and `Blue Pencil` toggles are imho redundant and might more confuse user than anything


Here is an example of working combo of settings (also note that enabling any of those two grease or blue pencil while disabling the HUD settings wont make any change and simply will have no effect)

Its kind of weird that even if set like below, it will still contain both pencils even those are disabled...

Image

Its not critical thing but very confusing to say the least...

And one last thing been that while using HUD and blue pencil it screws the thumbnail for some reason and having only pencils strokes visible but not the actual asset... probably for different PR tho or @BigRoy 

image

Its working good tho while inspecting the actual mp4 reviewable...which is good

image

moonyuet:
> Here is an example of working combo of settings (also note that enabling any of those two grease or blue pencil while disabling the HUD settings wont make any change and simply will have no effect

If turning on HUD means showing both grease pencil and blue pencil, shall we just mention in the docs or even in the descriptions to ensure users would turn on for these two?

> And one last thing been that while using HUD and blue pencil it screws the thumbnail for some reason and having only pencils strokes visible but not the actual asset... probably for different PR tho or @BigRoy

I believe this needs some settings change in extract thumbnail.

BigRoy:
> but it only works when those enabled so the `Grease Pencil` and `Blue Pencil` toggles are imho redundant and might more confuse user than anything

They are still relevant, you can enable HUD but still disable grease pencil/blue pencil. Because this:

> Its kind of weird that even if set like below, it will still contain both pencils even those are disabled...

Should indeed disable them. If it's not, that'd be a bug somewhere in our code I suspect. If I do the same in a Maya viewport it does keep HUD enabled and can hide e.g. the bluePencil just fine.

> And one last thing been that while using HUD and blue pencil it screws the thumbnail for some reason and having only pencils strokes visible but not the actual asset... 

---

> I believe this needs some settings change in extract thumbnail.

In the Maya one I suspect so, yes. Would be in this PR then.

client/ayon_maya/plugins/publish/extract_thumbnail.py
```diff
@@ -53,6 +53,9 @@ def process(self, instance):
             instance, camera, path,
             start=1, end=1,
             capture_preset=capture_preset)
+        if preset["viewport_options"].get("bluePencil"):
+            self.log.debug("Blue pencil is enabled, setting compression to jpg")
+            preset["compression"] = "jpg"
```
BigRoy, line 67: I'm not too keen on this - do we have slightly more information on WHY it fails with blue pencil? 🤔

moonyuet: it is submitting the transparent annotation instead of the bluepencil overlaying on the playblast images. I can open the thumbnail image in OpenRV and it shows exactly the same images as we have seen in the reviewables but not the thumbnail in the server.

moonyuet: [83431b2](https://github.com/ynput/ayon-maya/pull/465/commits/83431b2e82231cad862a135ede5396832d1aed3e) Updated the message.

BigRoy: Thanks, however - do we know **why** it separates the outlines? because in the playblasts it includes the model just fine. So something is wrong with the alpha when Blue pencil is enabled? 🤔 

It goes through the same way of rendering - does that mean the alpha is somehow wrong in the output image that has blue pencils (e.g. only blue pencil lines have alpha?)

moonyuet: > Thanks, however - do we know **why** it separates the outlines? because in the playblasts it includes the model just fine. So something is wrong with the alpha when Blue pencil is enabled? 🤔
> 
> It goes through the same way of rendering - does that mean the alpha is somehow wrong in the output image that has blue pencils (e.g. only blue pencil lines have alpha?)

Could be. It's quite odd though. But sometimes it is good to know: When I opened the thumbnail in png format via OpenRV, it shows. But it is a different story when it displays in the server.

https://github.com/user-attachments/assets/4c6b7676-0981-4feb-9c79-cb22ba4ce39a

BigRoy: My guess is that the thumbnail display we do pre-multiplies the image incorrectly 🤔 but that indeed the `alpha` of the blue pencil is white but the rest of the image is black (no alpha). Which can be confirmed by checking the alpha channel in RV I suppose?

moonyuet: You are right.
image

moonyuet: Added the additional comment to explain why the compression is there. [7dbc8b4](https://github.com/ynput/ayon-maya/pull/465/commits/7dbc8b427932e71a1e1536a30a872fb7ae4e82da)

client/ayon_maya/vendor/python/capture.py
```diff
@@ -219,7 +219,7 @@ def capture(camera=None,
                 )
                 stack.enter_context(_isolated_nodes(isolate, panel))
                 stack.enter_context(_maintained_time())
-
+                stack.enter_context(_enforce_blue_pencil_state(viewport_options))
```
BigRoy, line 223: Should probably also added this to the `if` clause abvoe?

client/ayon_maya/vendor/python/capture.py
```diff
@@ -12,7 +12,7 @@
 from maya import cmds
 from maya import mel
 
-from qtpy import QtGui, QtWidgets
+from qtpy import QtGui, QtWidgets, QtCore
```
BigRoy:
```suggestion
from qtpy import QtGui, QtWidgets
```

BigRoy (commented):
Overall looks ok to me; nice. @LiborBatek can you test some more?

LiborBatek:
@BigRoy I have tested all possible combo of those three settings and all good...now it behaves as expected

...also the thumbnails been correct and having what should be in there!

LGTM

